PR #11, opened 2023-01-23
massive refactor for speed.

laudiacay:
This reads the file from the remote disk once, encrypts and compresses in memory, then writes it to the local disk once.

It copies around within memory more than I'd like, but that's survivable.

It's about at the same level of code maturity as the rest of the repository.


amiller68 (commented):
Does this need `nightly`? Can you confirm / mark dependency in Readme?

amiller68:
Quick test showed we fixed the segfault, but no output is observed in target directory

amiller68 (approved):
Fine for now. Communicate use of nightly to x and t plz

PR #41, opened 2023-01-24
it writes to disk

laudiacay:


laudiacay:
this one is to add nice types that make sense and write out sensibly to a metadata file

PR #45, opened 2023-01-28
decryption

laudiacay:
- big refactor to prep for writing to output files and reconstructing
- we write the manifest to disk now
- doesn't compile / soz
- 99 previous errors on the wall, 99 previous errors. take one down, patch it around, 148 previous errors on the wall.
- compile 'er
- the start of a pipeline


laudiacay:
forcemerging so i can put thea's thing atop it and start testing (this is at the very least no *worse* than what is on master currently)

PR #46, opened 2023-01-28
this breaks things up into a workspace with some little things in it, as well as adds the first crappy little test (surprise! we are failing it)

laudiacay:


amiller68 (approved):
I almost cried from happiness. Approved

PR #52, opened 2023-02-01
Integration tests harness

ghost:


dataprep-pipelines/src/fsutil.rs
```diff
@@ -25,10 +25,11 @@ pub fn ensure_path_exists_and_is_empty_dir(path: &Path) -> Result<()> {
 // Note (amiller68): The following is not used.
 /*
    // TODO (laudiacay): Use proper function names
+   // TODO (thea-exe) look at this for tips
    // this comment lies in memoriam of the time i set these both to 10. if you estimate the disk
    // space used by a directory as only 512 bits, this would have filled 5 terabytes of disk space.
    // i'm not sure what i was thinking.
-   pub fn make_big_filesystem_clusterfuck(depth_to_go: usize, width: usize, cwd: PathBuf) {
+   pub fn make_big_filesystem_clusterfuck(depth_todepth_to_go: usize, width: usize, cwd: PathBuf_go: usize, width: usize, cwd: PathBuf) {
```
laudiacay: wanna remove this whole chunk from fsutil? make sure you're not deleting any necessary todos but we can probably get rid of it ifyou successfully integrated into ur tests

dataprep-pipelines/tests/integration.rs
```diff
@@ -1,44 +1,80 @@
...
 use dir_assert::assert_paths;
+use rand::Rng;
 use std::path::PathBuf;
 use tokio::fs;
+use tokio::io::AsyncWriteExt;
 
-#[tokio::test]
-async fn it_works_for_one_file() {
+lazy_static! {
+    static ref MANIFEST_FILE: PathBuf = PathBuf::from("test/manifest.json");
```
laudiacay: nice!! did you get how this uses types in a weird way to make this work? not necessary to use it correctly, but it's a really cool hack

dataprep-pipelines/tests/integration.rs
```diff
@@ -1,44 +1,80 @@
...
-    fs::create_dir("test/output").await.unwrap();
-    let output_dir = PathBuf::from("test/output");
+    fs::create_dir(&*OUTPUT_DIR).await.unwrap();
     // create final output directory for unpacked
-    fs::create_dir("test/unpacked").await.unwrap();
-    let unpacked_dir = PathBuf::from("test/unpacked");
-    let manifest_file = PathBuf::from("test/manifest.json");
-    // create a file in the input directory
-    fs::write("test/input/test.txt", b"test").await.unwrap();
+    fs::create_dir(&*UNPACKED_DIR).await.unwrap();
```
laudiacay: much cleaner

dataprep-pipelines/tests/integration.rs
```diff
@@ -1,44 +1,80 @@
...
+}
 
+async fn transform_and_check() {
     // run the function
     println!("doing pack pipeline!");
     dataprep_pipelines::pipeline::pack_pipeline::pack_pipeline(
-        input_dir.clone(),
-        output_dir.clone(),
-        manifest_file.clone(),
+        INPUT_DIR.clone(),
```
laudiacay: soon i'm going to ask you to remove like 90% of these pathbuf "clones" and make sure we're using all pointers to the same underlying data as much as possible

laudiacay: but not yet- can you make sure there's an issue for that though before we merge this?

dataprep-pipelines/tests/integration.rs
```diff
@@ -1,44 +1,80 @@
...
+
+async fn create_directory_structure(path: &str, length: usize) {
+    // creating a directory at the given path
+    fs::create_dir(path).await.unwrap();
+    // creates length number of directories and files within the directory at the given path
+    for i in 0..length {
+        // new directory path made
+        let dir_path = format!("{}/{}", path, i);
+        // create directory in the new directory path
+        fs::create_dir(dir_path.clone()).await.unwrap();
```
laudiacay: ok let's write some code that can make some more depths and widths of filesystem trees, different size files, really knock out those issues....

dataprep-pipelines/tests/integration.rs
```diff
@@ -1,44 +1,80 @@
...
+        let mut file = fs::File::create(file_path).await.unwrap();
+        // random data
+        let random_data = rand::thread_rng().gen::<[u8; 32]>();
+        file.write_all(&random_data);
+    }
+}
+#[tokio::test]
+async fn test_create_directory_structure() {
+    // assert and check the created directory structure
+    create_directory_structure("test_directory", 10);
```
laudiacay: so yeah now you need to run `transform_and_check` after create_directory_structure to make sure my code is doing the thing

laudiacay: also you need to call setup_structure before

and you need to ensure that create_directory is going into the proper `INPUT_DIR` subdirectory of `TEST_DIR`... does that make sense?

laudiacay (changes requested):

dataprep-pipelines/tests/integration.rs
amiller68: Taken care of

dataprep-pipelines/tests/integration.rs
amiller68: done

laudiacay (approved):
this is fine. the balanced is fine. i think you should screw the random for now and just make "very very long and skinny filesystem (dir dir dir dir dir dir dir dir dir file but with more dir's, this is to test for stack overflows on recursion)"
and then do "big honkin files all in one directory"
then do "a jillion tiny files all in one directory"
that should do for now

PR #131, opened 2023-03-01
Deduplication

recursivepaws:
This branch solves deduplication. Prior to this code implementation, duplicate files were being detected, but ignored in the packing step. As a result, filesystems with duplicate files were not able to be processed. This code addressed these issues by handling duplicate files, while only packing the original file. Test cases have been written to ensure this functionality.


recursivepaws:
the test failing here is `test_deduplication_size` which is designed to ensure packed filesystems with 50% duplicates are 50% of the footprint of filesystems with only unique files, *when they are packed*. This relies on an exact assertion that, while functional on my local machine, may not be exactly accurate in all environments. A better test condition would be to ensure that the packed dup filesystem is at least a certain factor smaller than its packed unique counterpart, rather than an exact factor. This can be updated in future versions.

dataprep-pipelines/src/plan_copy.rs
```diff
@@ -28,73 +28,102 @@ use crate::utils::hasher;
 pub async fn plan_copy(
     origin_data: SpiderMetadata,
     to_root: PathBuf,
-    seen_hashes: Arc<RwLock<HashMap<String, Rc<SpiderMetadata>>>>,
+    // This data structure is a little uglier than i would like it to be
+    seen_hashes: Arc<RwLock<HashMap<String, DuplicationMetadata>>>,
```
amiller68, line 32: Was this reference Counter important here?

amiller68: Tests pass so, i'm sure this is fine

amiller68 (dismissed):
This looks good!

amiller68 (approved):
Looks good!

PR #153, opened 2023-04-07
WNFS integration

recursivepaws:
# Description
Here, we gutted dataprep's packing and unpacking pipeline in favor of the utilization of WebNative FileSystem. 
Spidering still takes place, but many old structs have been removed and WNFS structs are used in their place.


amiller68 (dismissed):
we talked, performance is fine, happy to merge

laudiacay (approved):
fine once it's passing that one bug

laudiacay:
formatter is failing just kidding @organizedgrime can you fix that

PR #158, opened 2023-06-12
Pipeline Refactor

recursivepaws:
# Changes
- Refactored all occurrences of `dataprep` to `tomb`.
- Dramatically modified the project structure, separating our tool out into `tomb`, `tomb-common`, and `tomb-wasm` subcrates. 
- Pipelines now include `Add`, `Remove`, `Push`, and `Pull`, in addition to the preexisting `Pack` and `Unpack`. The way these pipelines are called is refactored. 
- Many functions have been modified to accept any type which conforms to the `BlockStore` trait, rather than one implementation of it in particular. 
- The `Pack` and `Unpack` pipelines can now be run directly on a remote endpoint via `NetworkBlockStore`, circumventing the need to build CAR files locally before syncing, as well as the need to download CAR files before reconstructing a given filesystem. 
- WNFS storage has been modified dramatically. There are now two `BlockStore`s and two `PrivateForest`s being used, as opposed to one of each. One Forest/Store pair encodes metadata about the filesystem structure, where the other Forest/Store pair encodes nothing but content data. 
- `Manifest`s have been modified to reflect this change in WNFS usage, and encode both a local and remote option for both the metadata `BlockStore` and the content `BlockStore`.
- Test cases now exist not only for the pipelines themselves but also for the CLI used to interact with them. 
- For all test cases to run correctly, they expect a [Kubo](https://github.com/ipfs/kubo) IPFS RPC node to be running locally via the command `ipfs daemon`, exposing port 5001. The presence of this daemon ensures our tools capacity to function with remote endpoints and `NetworkBlockStore` functionality.
- The serialization pipeline has been split into disk-based and disk-free functions, such that certain transformations can still be utilized in `tomb-common` / `tomb-wasm`.
- Serialization no longer relies on the `CarBlockStore` header to track root CIDs, instead tracking a `HashMap` of these CIDs separately. Doing so allows us to run these transformations with any `impl BlockStore` type, rather than CarBlockStore specifically, adds clarity to the utility of each object employed, and allows for more disk-free serialization. 
- The CLI now requires the initialization of a directory through the `tomb init` command to run pipelines on a given directory, and the subsequent configuration of remote endpoints through the `tomb config remote` command, if one wishes to sync their data remotely. 
- `tomb-wasm` has been set up to be the directory of our future web SDK, which we will be using in our frontend application. 

# Known TODOs
- XDG home directory conformance
    - Find more clever way to encode remote and local `BlockStore` info, maybe `Option`s
- Many `if local { <code> } else { <code> }` blocks could be simplified through the use of trait objects, but making `BlockStore` a trait which can be objectified may require further modification of [our rs-wnfs repository](https://github.com/banyancomputer/rs-wnfs). 
- [CarV2](https://ipld.io/specs/transport/car/carv2/) conformance
- Comprehensive WASM bindings and TypeScript functions for Banyan SDK
- `README.md` files ought to be present in each subcrate and helpful for both contributors and users. 
- Continuous Integration tests are still broken. Let's fix those!
- Make the metadata PrivateForest a subset of the content PrivateForest, which will contain both. This way, users that want to reconstruct their filesystems from Filecoin alone will not be adrift without access to our services.


.DS_Store
sstelfox, line 1: Ick hahah can we remove this and add a .gitignore entry to prevent these files from persisted in the repo?

sstelfox (dismissed):
Looks good to me 👍🏼

sstelfox (changes requested):
Looks like commitlint is sad with a couple of the commit messages in the branch. I only know how to change those en-masse using the CLI but your editor might be able to help. If you want to use the CLI running `git rebase -i` in the branch, finding the commits and change the word "pick" to "reword", saving and quitting should take you through modifying all the selected commits.

This will break the signatures on my commits but I can resign them.

PR #159, opened 2023-06-14
sstelfox/ENG-167: fix GitHub ci cargo in tomb

sstelfox:
# Description

This updates and sets a strong baseline for at least dependency monitoring with a stricter config. I had to disable the tests as they currently rely on remote services we don't have in the CI environment and likely need to be stubbed or pushed into the integration side of tests (which would be nice to split out into a separate test suite). This also loosens one of the commitlint rules that have been tripping up the CI in the past (not allowing capitals for any kind of subject such as Github). It's a dumb rule and I axed it.

This does push the testing down the road a little bit but we need to be able to merge PRs now. I'm not closing the ticket until the tests are also able to pass but we need to be able to merge again.

## Link to issue

* https://linear.app/banyan/issue/ENG-167/fix-github-ci-cargo-tests-in-tomb

## Type of change

## Test plan (required)

The ability to merge this PR is the test plan.


recursivepaws (approved):
This all looks good to me, lets get these merged! :)

PR #162, opened 2023-06-27
Vera / ENG-199 CARv2 conformance & Vera / ENG-215 XDG Home Standard

recursivepaws:
# Description
- `CarBlockStore` as it existed is now defunct, deprecated, and removed. 
    - In its place live `CarV1BlockStore` and `CarV2BlockStore`, which implement [CARv1](https://ipld.io/specs/transport/car/carv1/) and [CARv2](https://ipld.io/specs/transport/car/carv2/) standards respectively.
    - These `BlockStores` cover all cases presently, _except_ for indexed CARv2 files. Support for this can be added in the future, but the inbuilt CAR indexer is lightweight and efficient, skipping through the CAR to find CIDs and varints. This might not scale well, so we can fix this soon. 
- `Pack` and `Unpack` pipelines are no longer able to be run directly on a remote endpoint. Metadata and Content are required to be built and saved in a CAR files, and these CAR files must be built locally before being synced remotely. 
- `.tomb` folders are no longer generated in input, output, or reconstructed directories.
    - Instead, data regarding each "Bucket" / Root of a WNFS filesystem is tracked globally in a user's `$HOME/.config/tomb/global.json`
    - This `global.json` file tracks:
        - version of `tomb` used to create it
        - global remote endpoint configuration
        - bucket-specific configurations
            - the original path on disk of a filesystem
            - generated random folder name where packed data is stored for each bucket, stored in `$HOME/.local/share/tomb/<random folder name>`
            - paths of the `metadata.car` and `content.car` files contained within this random folder name
            - `root.key` is also stored in this generated folder, but its location is assumed / inferred and not stored directly in the config. 
    - This `global.json` file is loaded in on each run of each pipeline. It is used to look up metadata about a given origin path, if it exists. If it does not exist, it is created and serialized when the pipeline completes. 

## Link to issues

https://linear.app/banyan/issue/ENG-215/xdg-home-standard
https://linear.app/banyan/issue/ENG-199/carv2-conformance

## Type of change

- [x] This change requires a documentation update

## Test plan
- There are now 30 test cases for ensuring CAR file functionality. 
- Far fewer test cases are required to ensure the functionality of the new `GlobalConfig` and `BucketConfig` structs, considering how lightweight they are and how easily they serialize and deserialize, with the assistance of our serialization helper functions, for which test cases remain intact. 
- Many test cases were removed and/or deprecated given that we are no longer packing or unpacking directly to the network.
- `Pull` pipeline test cases are temporarily disabled, as my local IPFS instance inexplicably hangs on get_block requests, though this is not a problem with the test cases themselves.
- Because this new config paradigm requires that we use a global configuration file to track buckets, almost all of our test cases now require that we run them sequentially, rather than in parallel. Race conditions and locks are not accounted for on this global config file, and it is assumed there is only one writer at any given moment in time.


tomb/Cargo.toml
```diff
@@ -9,31 +9,31 @@ repository.workspace = true
 version.workspace = true
 
 [dependencies]
-tomb-common = { path = "../tomb-common", version = "1.1.0" }
+tomb-common = { path = "../tomb-common" }
```
sstelfox: While it is somewhat annoying to keep the version updated as we do releases, omitting it defaults to a wildcard which isn't accurate. The version needs to be present and tracking whatever the in-development version is currently listed in tomb-common's Cargo.toml file. This change is protected by cargo-deny (which is currently failing the PR because of this change).

tomb/src/lib.rs
```diff
@@ -35,22 +35,24 @@
...
-use serial_test as _;
+// use assert_cmd as _;
+// use clap as _;
+// use criterion as _;
+// use dir_assert as _;
+// use env_logger as _;
+// use fake_file as _;
+// use fs_extra as _;
+// use lazy_static as _;
+// use serial_test as _;
```
sstelfox: These should just be removed, they can always be added back if they need to become available.

tomb/src/pipelines/error.rs
```diff
@@ -0,0 +1,13 @@
+use thiserror::Error;
+
+/// Pipeline errors.
+#[derive(Debug, Error)]
+pub(crate) enum PipelineError {
+    #[error("Directory has not been initialized")]
+    Uninitialized(),
```
sstelfox: Remove the `()`

tomb/src/pipelines/error.rs
```diff
@@ -0,0 +1,13 @@
...
+    // #[error("Cannot find specified CID in block store: {0}")]
+    // CIDNotFound(Cid),
+
+    // #[error("Lock poisoned")]
+    // LockPoisoned,
```
sstelfox: We shouldn't leave unused commented-out code around, git allows us to look back at code that used to be there if its really necessary to bring back but usually its not needed in the future or can be trivially rewritten.

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
```diff
@@ -0,0 +1,311 @@
...
+    pub path: PathBuf,
+    pub(crate) carv1: CarV1,
+}
+
+impl CarV1BlockStore {
+    // Create a new CARv1 BlockStore from a file
+    pub fn new(path: &Path) -> Result<Self> {
+        // If the path is a directory
+        if path.is_dir() {
+            panic!("invalid path, must be file, not dir");
```
sstelfox: We definitely don't want to use `panic` anywhere that could be from a user error. This function is already returning an `anyhow::Error` so this for now could be as simple as `return Err("<existing msg>");`

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
```diff
@@ -0,0 +1,311 @@
...
+        else {
+            // Grab reader and writer
+            let mut w = car::get_write(path)?;
+            let mut r = car::get_read(path)?;
+
+            // Construct new
+            Ok(Self {
+                path: path.to_path_buf(),
+                carv1: CarV1::new(1, &mut r, &mut w)?
+            })
```
sstelfox: The way this is working seems a bit dangerous to me. The `crate::utils::car::*` methods here are opening two independent file descriptors to the same path that aren't aware of each other. If they are used at the same time, or the write is ever used before the read after the read has already been used you'll end up panic'ing due to an invalid seek on the file.

Writing is not appending, and will always start at the beginning unless this instance has already been written to since its creation. If something was reading from the file then its seek'd position will no longer exist. It's also possible that the writer will overwrite what the reader is trying to consume.

I haven't traced what its doing with the path being passed through to it yet, but to me that implies there might be _more_ file instances that could potentially end up reading or writing to that particular path independent of the already opened file descriptors which would potentially exacerbate this issue.

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
```diff
@@ -0,0 +1,311 @@
...
+        let tmp_file_name = format!(
+            "{}_tmp.car",
+            self.path.file_name().unwrap().to_str().unwrap()
+        );
+        let tmp_car_path = self.path.parent().unwrap().join(tmp_file_name);
+        let w = File::create(&tmp_car_path)?;
+        Ok((tmp_car_path, r, w))
+    }
+
+    fn tmp_finish(&self, tmp_car_path: PathBuf) -> Result<()> {
```
sstelfox: It looks like the only error types that can be generated here are `std::io::Error` we should be specific when possible

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
```diff
@@ -0,0 +1,311 @@
...
+    fn tmp_start(&self) -> Result<(PathBuf, File, File)> {
+        let r = car::get_read(&self.path)?;
+        let tmp_file_name = format!(
+            "{}_tmp.car",
+            self.path.file_name().unwrap().to_str().unwrap()
+        );
+        let tmp_car_path = self.path.parent().unwrap().join(tmp_file_name);
+        let w = File::create(&tmp_car_path)?;
+        Ok((tmp_car_path, r, w))
+    }
```
sstelfox: Similar to `tmp_finish` it looks like these are all `std::io::Error`

tomb-common/src/utils/car.rs
```diff
@@ -0,0 +1,12 @@
...
+    fs::{File, OpenOptions},
+    path::Path,
+};
+
+pub fn get_read(path: &Path) -> Result<File> {
+    Ok(OpenOptions::new().read(true).open(path)?)
+}
+pub fn get_write(path: &Path) -> Result<File> {
+    Ok(OpenOptions::new().append(false).write(true).open(path)?)
+}
```
sstelfox: Both of these can only generate `std::io::Error` so that should be used as part of the Result type. I'm not entirely sure these helper methods are useful, elsewhere in the code these are used with comments assuming `get_write` is append-only which isn't the case here. This also doesn't give us the flexibility of not using files which seems like a pretty big limitation.

tomb-common/src/types/pipeline.rs
```diff
@@ -1,46 +1,35 @@
...
-            .finish()
-    }
-}
+// impl Debug for Manifest {
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         f.debug_struct("Manifest")
+//             .field("version", &self.version)
+//             .finish()
+//     }
+// }
```
sstelfox: We don't need to keep around commented out code

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
```diff
@@ -0,0 +1,311 @@
...
+        let block = V1Block::new(bytes, codec)?;
+        // If this CID already exists in the store
+        if self.get_block(&block.cid).await.is_ok() {
+            // Return OK
+            Ok(block.cid)
+        }
+        // If this needs to be appended to the CARv1
+        else {
+            // Open the file in append mode
+            let mut file = car::get_write(&self.path)?;
```
sstelfox: This is one of the cases where the use of the helper method is incorrect. `get_write` _does not_ open the file in append mode, it will rewrite the entire contents of the file the way this currently work.

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
```diff
@@ -0,0 +1,311 @@
...
+        }
+    }
+}
+
+impl Serialize for CarV1BlockStore {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        self.to_disk().unwrap();
```
sstelfox: We shouldn't be unwrapping pretty much anywhere, the Serialize result naturally allows for `std::io::Errror` to happen so the error type can be appropriately returned here.

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
```diff
@@ -0,0 +1,311 @@
...
+        self.path.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for CarV1BlockStore {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Ok(Self::new(&PathBuf::deserialize(deserializer)?).unwrap())
```
sstelfox: Similar to the serializer we shouldn't unwrap and instead return the error.

tomb-common/src/types/blockstore/car/carv1/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+
+        // Skip to the point where the old data started
+        r.seek(SeekFrom::Start(
+            carv1_start + *self.read_header_len.borrow(),
+        ))?;
+
+        // Whiel we're able to successfully read in blocks
+        while let Ok(block_offset) = r.stream_position() &&
+              let Ok((varint, cid)) = V1Block::start_read(&mut r) &&
+              let Ok(block) = V1Block::finish_read(varint, cid, &mut r) {
```
sstelfox: This let syntax is really weird and definitely shouldn't be chained like this. Based on what it's doing this probably should be checking whether it's at the end of the file, and if any error occurs and it's _not_ at the end of the file, probably means the file is corrupt and we shouldn't be writing to it.

tomb-common/src/types/blockstore/car/carv1/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// Write out a CARv1 object, assuming the Writer is already seeked to the first byte of the CARv1
+    pub(crate) fn write_bytes<R: Read + Seek, W: Write + Seek>(
+        &self,
+        mut r: R,
+        mut w: W,
+    ) -> Result<()> {
+        // Save our starting point
+        let carv1_start = r.stream_position()?;
+        w.seek(SeekFrom::Start(carv1_start))?;
```
sstelfox, line 53: A lot of the seeking while writing concerns me, if it's an existing file and we write somewhere in the middle of the file, we will inevitably corrupt the file unless the bytes being written are exactly the same length as an existing block record. If its shorter we'll be leaving behind orphaned garbage data, if longer we'll be corrupting the next (and possibly more) blocks.

tomb-common/src/types/blockstore/car/carv1/v1block.rs
```diff
@@ -0,0 +1,105 @@
...
+        let cid_buf: Vec<u8> = self.cid.to_bytes();
+        // Write all bytes
+        w.write_all(&varint_buf)?;
+        w.write_all(&cid_buf)?;
+        w.write_all(&self.content)?;
+        // Flush
+        w.flush()?;
+        // Return size
+        Ok(varint_buf.len() + cid_buf.len() + self.content.len())
+    }
```
sstelfox: It seems like we're doing a lot of CBOR encoding effectively by hand. Is there any particular reason we're not using an existing library for these records like `ciborium`?

tomb-common/src/types/blockstore/car/carv1/v1header.rs
```diff
@@ -0,0 +1,187 @@
...
+            .borrow()
+            .iter()
+            .map(|&root| Ipld::Link(root))
+            .collect();
+        // Insert the roots into the map
+        map.insert("roots".to_string(), Ipld::List(ipld_roots));
+        // Construct the final IPLD
+        let ipld = Ipld::Map(map);
+        dagcbor::encode(&ipld)
+    }
```
sstelfox: This looks like it might be conflating parts of the [CARv2 spec](https://ipld.io/specs/transport/car/carv2/) with [CARv1](https://ipld.io/specs/transport/car/carv1/#format-description). CARv1 (this file) only encodes a varint version number (which is always 1 for CARv1) and then the Ipld::List containing the roots. The Map here that contains the version is part of the fixed CARv2 header.

tomb-common/src/types/blockstore/car/carv1/v1block.rs
```diff
@@ -0,0 +1,105 @@
...
+use anyhow::Result;
+use std::io::{Read, Seek, Write};
+use wnfs::libipld::{
+    multihash::{Code, MultihashDigest},
+    Cid, IpldCodec,
+};
+
+// | 19-byte varint | x-byte Cid | x-byte content |
+#[derive(PartialEq, Debug)]
+pub(crate) struct V1Block {
```
sstelfox: This struct is already versioned by its module path. I'd leave off the V1 here. If it needs to be imported in a place with a competing version of the name it can be imported with a rename to disambiguate it such as.

tomb-common/src/types/blockstore/car/carv1/v1header.rs
```diff
@@ -0,0 +1,187 @@
...
+    io::{Read, Seek, Write},
+};
+use wnfs::{
+    common::dagcbor,
+    libipld::{Cid, Ipld},
+};
+
+// | 16-byte varint | n-byte DAG CBOR |
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+pub(crate) struct V1Header {
```
sstelfox: This struct is already versioned by its module path. I'd leave off the V1 here. If it needs to be imported in a place with a competing version of the name it can be imported with a rename to disambiguate it such as.

tomb-common/src/types/blockstore/car/carv1/v1index.rs
```diff
@@ -0,0 +1,126 @@
...
+use serde::{Deserialize, Deserializer, Serialize};
+use std::{
+    collections::HashMap,
+    io::{Read, Seek, SeekFrom},
+    str::FromStr,
+};
+use wnfs::{common::BlockStoreError, libipld::Cid};
+
+#[derive(Debug, PartialEq, Default, Clone)]
+pub struct V1Index {
```
sstelfox: This struct is already versioned by its module path. I'd leave off the V1 here. If it needs to be imported in a place with a competing version of the name it can be imported with a rename to disambiguate it such as.

tomb-common/src/types/blockstore/car/carv2/carv2blockstore.rs
```diff
@@ -0,0 +1,228 @@
...
+    fs::{remove_file, rename, File},
+    path::{Path, PathBuf},
+};
+use wnfs::{
+    common::BlockStore,
+    libipld::{Cid, IpldCodec},
+};
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct CarV2BlockStore {
```
sstelfox: This struct is already versioned by its module path. I'd leave off the CarV2 here. If it needs to be imported in a place with a competing version of the name it can be imported with a rename to disambiguate it such as.

tomb-common/src/types/blockstore/car/carv2/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+pub(crate) const V2_PRAGMA_SIZE: usize = 11;
+pub(crate) const V2_PH_SIZE: u64 = 51;
+
+// This is the fixed file signature associated with the CARV2 file format
+pub(crate) const V2_PRAGMA: [u8; V2_PRAGMA_SIZE] = [
+    0x0a, 0xa1, 0x67, 0x76, 0x65, 0x72, 0x73, 0x69, 0x6f, 0x6e, 0x02,
+];
+
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+pub struct CarV2 {
```
sstelfox: The V2 here is implied through the module path. I would leave off the specific identifiers. This is commonly a best practice in Rust and something clippy will generally complain about.

tomb-common/src/types/blockstore/car/carv2/v2header.rs
```diff
@@ -0,0 +1,143 @@
...
+
+use crate::types::blockstore::car::varint::{
+    encode_varint_u128_exact, encode_varint_u64_exact, read_varint_u128_exact,
+    read_varint_u64_exact,
+};
+
+pub const V2_HEADER_SIZE: usize = 40;
+
+// | 16-byte characteristics | 8-byte data offset | 8-byte data size | 8-byte index offset |
+#[derive(Debug, PartialEq, Clone, Copy)]
```
sstelfox: Mentioned this a couple of times of but V2 prefix is implied through the module path and best practice is to leave it out (allowing renaming of imports for disambiguating competing versions.

tomb-common/src/types/blockstore/car/carv2/carv2blockstore.rs
```diff
@@ -0,0 +1,228 @@
...
+            let mut r = car::get_read(path)?;
+            // Create new 
+            let store = CarV2BlockStore {
+                path: path.to_path_buf(),
+                carv2: CarV2::new(&mut r, &mut w)?
+            };
+            // Return Ok
+            Ok(store)
+        }
+    }
```
sstelfox: This looks like duplicated code, and I had some serious concerns about the V1 version. See the notes there.

tomb-common/src/types/blockstore/car/carv2/carv2blockstore.rs
```diff
@@ -0,0 +1,228 @@
...
+    path::{Path, PathBuf},
+};
+use wnfs::{
+    common::BlockStore,
+    libipld::{Cid, IpldCodec},
+};
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct CarV2BlockStore {
+    pub path: PathBuf,
```
sstelfox: Both versions are tracking the path, but this seems like an impediment to a more general storage system. We don't know the car needs to be written to a file on disk necessarily. I know this is used later on with the utils::* methods but I think those are problematic (see the notes I left in the CARv1 version).

tomb-common/src/types/blockstore/car/carv2/carv2blockstore.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl<'de> Deserialize<'de> for CarV2BlockStore {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        Ok(Self::new(&PathBuf::deserialize(deserializer)?).unwrap())
+    }
+}
```
sstelfox: See my related comments in the V1 version

tomb-common/src/types/blockstore/car/carv2/v2index.rs
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    // pub fn write_bytes<W: Write + Seek>(&self, mut w: W) -> Result<()> {
+    //     // Write codec
+    //     w.write_all(&encode_varint_u128(self.codec))?;
+    //     // Write bytes
+    //     w.write_all(&self.bytes)?;
+    //     Ok(())
+    // }
+}
```
sstelfox: After scanning the V2 docs it looks like we are going to want an index for the V2. The V1 spec strongly recommends against including more than 1 root with that spec alone and scanning the WNFS spec it seems that each filesystem entry should be its own "root" data block inside the CAR file. We'll want to be able to look these up without scanning the entire file which requires some form of index. Since we want these CAR files to be largely self-contained that means we need an index.

tomb-common/src/types/blockstore/car/carv2/v2header.rs
```diff
@@ -0,0 +1,143 @@
...
+impl<'de> Deserialize<'de> for V2Header {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let header_bytes = <Vec<u8>>::deserialize(deserializer)?;
+        let new_header = Self::read_bytes(header_bytes.as_slice()).unwrap();
+        Ok(new_header)
+    }
+}
```
sstelfox: We shouldn't be unwrapping in these

tomb-common/src/types/blockstore/car/carv1/v1index.rs
```diff
@@ -0,0 +1,126 @@
...
+              let Ok((varint, cid)) = V1Block::start_read(&mut r) {
+            // Log where we found this block
+            map.insert(cid, block_offset);
+            // Skip the rest of the block
+            r.seek(SeekFrom::Current(varint as i64 - cid.to_bytes().len() as i64))?;
+            next_block = r.stream_position()?;
+        }
+
+        Ok(Self { map, next_block })
+    }
```
sstelfox: We really don't want to scan entire files everytime we need to inspect a CARv1 file. We should probably ditch the V1 index entirely and use the V2 index to locate blocks within the CAR file

tomb-common/src/utils/config.rs
```diff
@@ -1,28 +1,28 @@
...
+const HOME_ERROR: &str = "cant find home directory";
 
 /// Grab config path from env variables + XDG spec
-pub fn tomb_config() -> Result<PathBuf> {
+pub fn xdg_config_home() -> PathBuf {
     // Construct
-    let path = PathBuf::from(format!("{}/.config/tomb", env::var("HOME")?));
+    let path = PathBuf::from(format!(
+        "{}/.config/tomb",
+        env::var("HOME").expect(HOME_ERROR)
```
sstelfox, line 10: This won't work for windows machines but I think that's fine for now. We just need to be aware of the limitation.

tomb-common/src/utils/serialize.rs
```diff
@@ -1,15 +1,18 @@
...
     libipld::{serde as ipld_serde, Cid, Ipld},
     private::{PrivateDirectory, PrivateForest, PrivateNode, PrivateRef, TemporalKey},
 };
 
 /// Store a given PrivateForest in a given Store
-pub async fn store_forest(forest: &Rc<PrivateForest>, store: &impl BlockStore) -> Result<Cid> {
+pub(crate) async fn store_forest(
+    forest: &Rc<PrivateForest>,
+    store: &impl BlockStore,
+) -> Result<Cid> {
```
sstelfox: This pattern of creating a function for each type isn't ideal. Is there instead a trait that could be used for storing each of these entries in the blockstore implementation?

tomb-wasm/src/types/pipeline.rs
```diff
@@ -1,6 +1,6 @@
...
-use wasm_bindgen::prelude::wasm_bindgen;
+// use tomb_common::types::pipeline::Manifest as TombManifeset;
+// use wasm_bindgen::prelude::wasm_bindgen;
 
-#[wasm_bindgen]
-#[derive(Default)]
-pub struct Manifest(pub(crate) TombManifeset);
+// #[wasm_bindgen]
+// #[derive(Default)]
+// pub struct Manifest(pub(crate) TombManifeset);
```
sstelfox: We don't need to keep commented out code kicking around

tomb-wasm/src/utils/metadata.rs
```diff
@@ -1,21 +1,21 @@
...
-// Provide a default manifest
-#[wasm_bindgen]
-pub async fn default_manifest() -> Result<Manifest, JsError> {
-    Ok(Manifest::default())
-}
+// // Provide a default manifest
+// #[wasm_bindgen]
+// pub async fn default_manifest() -> Result<Manifest, JsError> {
+//     Ok(Manifest::default())
+// }
```
sstelfox: We don't need to keep commented out code (or uses at the top)

tomb/src/main.rs
```diff
@@ -52,53 +33,51 @@ async fn main() -> Result<()> {
...
-                Some(output_dir) => {
-                    pack::pipeline(&input_dir, Some(&output_dir), chunk_size, follow_links).await?;
-                },
-                None => {
-                    pack::pipeline(&input_dir, None, chunk_size, follow_links).await?;
-                },
+            if let Some(input_dir) = input_dir {
+                pack::pipeline(&input_dir, follow_links).await?;
+            }
+            else {
```
sstelfox: Little nitpick but same line should be preferred here (cargo fmt will fixup all these kinds of things)

tomb/src/pipelines/pack.rs
```diff
@@ -48,182 +42,75 @@ use wnfs::{
...
+        // if first_run {
+        //     println!("storing original dir and key");
+        //     let original_key = store_dir(
+        //         &mut manifest,
+        //         &mut metadata_forest,
+        //         &root_dir,
+        //     )
+        //     .await?;
+        //     key_to_disk(tomb_path, &original_key, "original")?;
+        // }
```
sstelfox: We don't need to keep commented out code kicking around

tomb/src/utils/tests.rs
```diff
@@ -1,38 +1,37 @@
...
     ensure_path_exists_and_is_empty_dir(&root_path, true)?;
-    // Input and output paths
+    // Input and path
     let input_path = root_path.join("input");
-    let output_path = root_path.join("output");
-    create_dir_all(&output_path)?;
     // Generate file structure
     structure.generate(&input_path)?;
+    // Deinitialize existing data / metadata
+    // configure::deinit(&input_path)?;
```
sstelfox: We don't need to keep commented out code kicking around

sstelfox (changes requested):
Sorry about the number of comments, I went through everything in here even though a lot of it seems like preexisting code.

There are a couple of failing CI checks as well, and I'd generally recommend you run commits through `cargo fmt` and `cargo clippy` before committing (a pre-commit hook can help here) as I think they'd catch a few things I commented on. Before you request a code review on a PR please check and make sure the existing CI tests are passing.

I also know that some of what I pointed out are larger changes to the code base but I think they're all important.

tomb/src/lib.rs
recursivepaws: my bad, i removed the lib.rs requirement that all libraries be used, and commented these out while i sorted things into testing and prod dependencies. will update

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
recursivepaws: Replacing this with a CarError type!

tomb-common/src/utils/car.rs
recursivepaws: i agree that these methods arent particularly helpful, they're only used in the context of FileIO-based CAR blockstores. will see if there is a better place to put them, or if they can be removed entirely. Only being reused in CarV1BlockStore and CarV2BlockStore, and definitely wont make it anywhere else.

tomb-common/src/types/blockstore/car/carv2/carv2blockstore.rs
recursivepaws: it's my understanding that we will need to reexport the CarV2BlockStore and CarV1BlockStore types for WASM anyway, which is why i've tried to keep these files comparatively barebones when contrasted with the Car types in each mod.rs, but I agree that we want to minimize as much code repetition as possible. We can discuss this further

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
recursivepaws: you're correct that it's not being opened in append mode even though it should be here, but why would this cause us to rewrite the entire file contents? arent we still Seeking to the next_block location before writing anything?

tomb-common/src/types/blockstore/car/carv1/mod.rs
recursivepaws: i'll simplify this a bit, but there's an important distinction here which is that not finding another block doesn't necessarily mean we are at the end of the file, the CarV2 index might still succeed it. Therefore we should try to read these in as blocks until we fail to do so, or, if we want to be safer, also include a special stop case here for when the CarV2 header has an index_offset != 0. 
i would suggest that we implement separate functions for ensuring the integrity / non-corrupt status of a given file, which we can run before doing important operations like saving to disk.

tomb-common/src/utils/serialize.rs
recursivepaws: this will be superceded by single root CID metadata IPLD format discussed during standup today

tomb-common/src/types/blockstore/car/carv1/mod.rs
recursivepaws: i believe that this is something we can fix by traversing backwards through the file. if we already have an map of blocks and their offsets built, and start with the blocks written furthest into the file, then there is no chance of overwriting existing data. either the header has increased in size or remained fixed in size. in the former case, all blocks are being shifted further into the file, so reading a 100 byte block and skipping 50 bytes of offset into it to overwrite it wont matter if there are no blocks succeeding it. in the latter case, no writing even needs to be performed. 
that might still sound unsafe to you, but theres no longer a chance of destroying block data, even if we corrupt the file format. 
we could likely implement this with the existing v1 index, but id rather wait until we have a v2 index parser, as it will make the v1 index essentially obsolete for our use case.

tomb/Cargo.toml
recursivepaws: fixing

tomb-common/src/types/pipeline.rs
recursivepaws: removed 👍🏻

tomb-common/src/types/blockstore/car/carv1/v1header.rs
recursivepaws: The [CARv1 header specification](https://ipld.io/specs/transport/car/carv1/#header) does imply that both the version number and roots list should be present in this IPLD map. This version is successfully deserialized in the car fixtures we are provided. Additionally, based on my analysis of the way fixtures deserialize, this version is actually set to `2` if we are enclosed in a CARv2. So I believe this code is actually correct and necessary.

tomb-wasm/src/utils/metadata.rs
recursivepaws: removed

tomb-common/src/types/blockstore/car/carv2/carv2blockstore.rs
recursivepaws: corrected

tomb-common/src/types/blockstore/car/carv2/v2header.rs
recursivepaws: corrected

tomb-common/src/types/blockstore/car/carv1/v1block.rs
recursivepaws: wnfs already has a `cbor::{encode, decode}` functionality that we use elsewhere in tomb, but varint code specifically i implemented myself because it requires byte level precision and can come in two forms. the V1 varints minimize bytes, but the V2 header varints are of deterministic length. Happy to utilize a library for this or the other reads in this snippet if one is deemed appropriate, but for the time being i believe this is sufficient.

tomb-common/src/types/blockstore/car/carv2/v2index.rs
recursivepaws: agreed; i've known and acknowledged the v2index will be required at scale- just hasn't been a priority yet. will have to ticket this separately.

tomb/src/pipelines/unpack.rs
```diff
@@ -18,117 +20,95 @@ use wnfs::{
...
+            "🔐 Decompressing and decrypting each file as it is copied to the new filesystem at {}",
+            output_dir.display()
+        );
 
-                // For each of those children
-                for node_name in node_names {
-                    // Fetch the Node with the given name
-                    if let Some(node) = dir
-                        .get_node(&[node_name.clone()], true, hot_forest, hot_store)
+        #[async_recursion(?Send)]
```
sstelfox, line 40: Huh I've never seen this trait before, neat

sstelfox (dismissed):
Very nice! Let's get this merged 🎉

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
sstelfox: Yeah seeks solve the issue, its just unusual to see starting the seek cursor at the beginning of a file unless the intention is to overwrite it entirely.

tomb-common/src/types/blockstore/car/carv1/mod.rs
sstelfox: Well we should generally reject invalid CAR files, and those cases I covered mean the data is either corrupt or malicious at some level. I'm more concerned about reading in strange car files (such as an import, or someone pushing to our service using a third party client). We should only be producing valid files, if we're not that's going to lead to corruption eventually somewhere important.

sstelfox: Its something I think we need to fix but it doesn't have to be right now.

tomb-common/src/types/blockstore/car/carv1/carv1blockstore.rs
sstelfox: From out of band: We're going to address this in a future PR

amiller68 (dismissed):
Looks great!
Left u a couple conflicts from my PR, let me know if u want help resolving them and then we can get this merged

PR #163, opened 2023-06-28
Eng-212: Local Device Key Generation

amiller68:
# Description

Re-opened in new Repo as per request in: https://github.com/banyancomputer/rs-wnfs/pull/8

Move's work done on that PR into Tomb directly.


tomb-common/src/crypto/rsa.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    async fn from_modulus(modulus: &[u8]) -> Result<Self> {
+        let n = BigUint::from_bytes_le(modulus);
+        let e = BigUint::from(PUBLIC_KEY_EXPONENT);
+
+        Ok(Self(
+            rsa::RsaPublicKey::new(n, e).map_err(|e| RsaError::InvalidPublicKey(anyhow!(e)))?,
+        ))
+    }
+}
```
sstelfox, line 162: Does this actually get used anywhere other than in the test for it? If not we probably don't want it.

sstelfox (changes requested):
One more open question and a failing CI check but otherwise looks good to me 👍🏼

tomb-common/src/crypto/rsa.rs
amiller68: This is required for implementing `ExchangKey` 🤷‍♂️

sstelfox (dismissed):

tomb-common/src/crypto/rsa.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    /// Reads the private key from DER bytes.
+    ///
+    /// # Arguments
+    /// bytes - The DER bytes to read from.
+    pub fn from_der(bytes: &[u8]) -> Result<Self> {
+        let key = rsa::RsaPrivateKey::from_pkcs8_der(bytes)?;
+        Ok(Self(key))
+    }
+
```
recursivepaws, line 137: id like that there be a `to_der()` function exposed here as well

recursivepaws (changes requested):
These changes all look good to merge, but would like `to_der` fn given that the internal type wont be exposed

tomb-common/src/crypto/rsa.rs
amiller68: I can implement this -- but I am wondering what use case there is for exporting key bytes to der
- this isn't a key you share
- there's already an export method (to_pem_file)

recursivepaws: i believe we will want to be able to store the public keys in the metadata CAR file, which means serializing and deserializing them without writing them to disk as PEMs

recursivepaws (dismissed):
Looks good! Let's merge

sstelfox (approved):
Look good to me 👍🏼

recursivepaws (approved):
👍🏻

PR #166, opened 2023-07-06
Vera/ENG-224 Temporal Key Shadowing

recursivepaws:
# Description
- `TemporalKey`s as described by `rs-wnfs` are no longer stored directly on disk. Rather, `RsaPublicKey`s are utilized to encrypt those `TemporalKey`s. 
- Using the RSA key's fingerprint as a `HashMap` key, corresponding `RsaPublicKey`s and encrypted `TemporalKeys` are tracked in a key `Manager` and `Mapper`s.
- This `Manager` struct is serialized in the Metadata `BlockStore`, allowing us to recover `Mapper`s for both the original `TemporalKey` and the most recent `TemporalKey`. We store both for future use in versioning.
- With the presence of any of the `RsaPrivateKey`s that correspond to inserted `RsaPublicKey`s, the original and current `TemporalKey`s can be reconstructed, allowing for the decryption and modification of the existing WNFS filesystem.
- CAR-based `BlockStore`s are now implementing a single root for our use case
    - This single root contains an `IPLD::Map` which has keys corresponding to the `CID`s of all relevant structs:
        - `metadata_forest`
        - `content_forest`
        - `current_private_ref`
        - `original_private_ref`
        - `key_manager`
    - This root is stored both in our metadata `BlockStore` and our content `BlockStore`, though currently all data is not duplicated across both. 
- Moved `create_plans` and `process_plans` functions into their own separate `utils/pack.rs` file to keep the actual `pipelines/pack.rs` file more succinct. 
- `input_dir` has been renamed to `origin` in various locations to maintain congruency. 
    - We are now thinking of `Bucket`s in terms of which directory is its 'origin'. Metadata and Content CAR files can be looked up about this `Bucket` through use of its origin in the CLI, even if the directory referenced no longer exists, because the `GlobalConfig` searches for `Bucket`s in its `JSON` using the `Path` in which the original filesystem existed. 

## Link to issue
[ENG-224 Temporal Key Shadowing](https://linear.app/banyan/issue/ENG-224/temporal-key-shadowing)

## Type of change

## Test plan (required)
- Renamed myriad test cases to reduce redundancy in naming.
- GlobalConfig utilizes `to_from_disk`, `add_bucket`, and `from_disk_direct` to ensure it functions as expected. 
- `put_get_original`, `put_get_current`, `insert_get_current`, `insert_get_original`, and `insert_get_both` test cases exist in the mod.rs file for the keys submodule. These test cases ensure: 
    - Updating the `TemporalKey` updates all encrypted `TemporalKey`s.
    - Inserting a new `RsaPublicKey` generates a new encrypted `TemporalKey` in the `Mapper`.
    - `TemporalKey`s can be decrypted with the use of `RsaPrivateKey`s both before and after modification.
- The `utils::serialize::test::all` test now includes an assertion for ensuring the key `Manager` is successfully reconstructed. This ensures the consistency outlined in other test cases is maintained before and after `BlockStore` serialization.
- The new method of serializing data using a single root `IPLD::Map` is verified using the tests `utils::serialize::test::{forest, all, dir_content, dir_object, history}`


tomb-common/src/types/config/keys/mapper.rs
```diff
@@ -0,0 +1,75 @@
...
+        Ok(())
+    }
+
+    pub async fn insert_public_key(
+        &mut self,
+        temporal_key: &Option<TemporalKey>,
+        new_key: &RsaPublicKey,
+    ) -> Result<()> {
+        // Grab the public key's fingerprint
+        let fingerprint = hex::encode(new_key.get_fingerprint()?);
```
amiller68: This is already a hex encoded string :)

tomb-common/src/types/config/keys/mapper.rs
```diff
@@ -0,0 +1,75 @@
...
+            self.0.insert(fingerprint, (der, vec![]));
+        }
+
+        // Return Ok
+        Ok(())
+    }
+
+    pub async fn reconstruct(&self, private_key: &RsaPrivateKey) -> Result<TemporalKey> {
+        // Grab the fingerprint
+        let fingerprint = hex::encode(private_key.get_public_key().get_fingerprint()?);
```
amiller68: Same here -- no need to hex::encode

tomb/src/utils/pack.rs
```diff
@@ -0,0 +1,199 @@
...
+                            Utc::now(),
+                            metadata_forest,
+                            metadata,
+                            rng,
+                        )
+                        .await?;
+                }
+                // We don't need an else here, directories don't actually contain any data
+            }
+            PackPipelinePlan::Symlink(_, _) => todo!(),
```
amiller68: Not sure if this is ever matched

tomb/src/utils/pack.rs
```diff
@@ -0,0 +1,199 @@
...
+                        &symlink_segments,
+                        true,
+                        Utc::now(),
+                        metadata_forest,
+                        metadata,
+                        rng,
+                    )
+                    .await?;
+            }
+            PackPipelinePlan::Directory(_) | PackPipelinePlan::FileGroup(_) => todo!(),
```
amiller68: Not sure if this ever matched

amiller68 (changes requested):
Looks good! Just some stuff I was confused about

tomb/src/utils/pack.rs
amiller68: is anything -- change to explicit panic

tomb/src/utils/pack.rs
amiller68: is anything -- change to explicit panic

amiller68 (approved):

PR #168, opened 2023-07-11
Vera/ENG-39 Tomb WASM Binding

recursivepaws:
# Description
- Bug fixes have been applied to the `MultifileBlockStore` and its expected behavior. 
    - Specifically, if a user deserializes a `MultifileBlockStore` but does not write to it, the new delta file created is not saved. 
- `tomb-wasm` sub-crate now includes:
    - functionality for loading in a `WasmBlockStore` from a remote url of a `CARv2`
    - functionality for loading in a `PrivateKey` from a remote url of a `.pem`
    - functionality for loading in a `TombFS` from both a `WasmBlockStore` and a `PrivateKey`
        - This `TombFS` exposes an `ls` function which accepts a JS array of path strings, and returns a `JavaScript` array object, where each item in the array contains both a `file_name` and `metadata` field, which can be queried by the browser. 
        - With these methods exposed, we now meet the MVP requirements for a WASM crate which can interact with our existing infrastructure.

## Link to issue
[Vera/ENG-39 Tomb WASM Binding](https://linear.app/banyan/issue/ENG-39/tomb-wasm-binding)

## Type of change

## Test plan (required)
- Testing on WASM TombFS functionality is being performed using the files hosted at [this Gist](https://gist.github.com/organizedgrime/f292f28a6ea39cea5fd1b844c51da4fb). They encode a `meta.car` file, a `content.car` file, and the `wrapping_key.pem` that was used to create them with the most recent version of `tomb`. 
- Using these files allows us to verify that decryption can be run in browser as it would be in the CLI.


tomb-wasm/src/metadata/metadata.rs
sstelfox, line 1: Probably don't need this empty file hahah. It doesn't look like its referenced as a module either.

sstelfox (dismissed):
There is one extra file that probably doesn't need to be there but otherwise seems good 👍🏼

sstelfox (dismissed):

sstelfox (approved):

PR #170, opened 2023-07-14
Vera/Documentation & Cleanup

recursivepaws:
# Description
- Added a great deal of documentation, refactored a few struct names and moved a few things around to reduce naming redundancy.
- Removed old FileSystem code from `tomb-wasm` which is no longer congruent with `TombFS`
- Separated out the CAR errors from the IO errors that occur when dealing with their files.

## Type of change
- [x] Comments have been added/updated.


tomb-wasm/src/metadata/crypto.rs
```diff
@@ -4,6 +4,7 @@ use wasm_bindgen::{prelude::wasm_bindgen, JsValue};
 use crate::fetch::http::get_data;
 
 #[wasm_bindgen]
+#[derive(Debug)]
```
sstelfox: We probably don't want private keys to allow any form of printable output (even Debug). It's just a safety precaution against these values showing up in any logs or diagnostic output. If this was added because it a type embeds it and is also implementing Debug, you'll probably need to create a custom Debug implementation to omit the field but that's really not too bad.

sstelfox (changes requested):
Saw one thing but mostly looks good 👍🏼

sstelfox (approved):

PR #171, opened 2023-07-14
Bring in the EC key sealing code

sstelfox:
Still need to add better error handling, tests, and docs but the public type interfaces should all be stable.


recursivepaws (approved):
Look good to me. Will refactor and add lints and error handling once this is merged

PR #177, opened 2023-07-14
Move all the non-wasm code into its own little world

sstelfox:
* Reorganized types into their own files
* Created crate internal trait to ensure consistency between target implementations
* Moved standard openssl based workflow aside to support an independent wasm version

Still have outstanding tasks that will be coming in another PR:

* Documentation!
* Tests!
* More Reorganization


amiller68 (dismissed):
looks gud to me

amiller68 (approved):
this allllsoooop luks gud

PR #183, opened 2023-07-18
Alex/eng 258 implement wasm ecdsaecdh key handling for tomb

amiller68:
# THIS IS A DRAFT

# Description

Implements parity of features from Tomb-Crypt in wasm


sstelfox (approved):
Went through this with Alex in a pair session. Move the library into its own available [here](https://github.com/banyancomputer/tomb-crypt). This PR will remove the duplicate code and point the dependencies to the correct version.

PR #186, opened 2023-07-20
Vera/ENG-260 Car Stream Safety

recursivepaws:
# Description
- `CARv1` and `CARv2` objects were previously being loaded in using both a `R: Read + Seek` and a `W: Write + Seek`. 
    - In some cases, the two streams were expected to reference the same content, while in other cases they were not. 
    - This behavior was definitionally unsafe.
- Cases where both a reader and writer were used have now been replaced by a single generic `RW: Read + Write + Seek`.
    -  This ensures the safety of file descriptors and prevents us from overwriting data in place. 
- Additionally, the `CARv1` and `CARv2` that builds on top of it can now be safely rewritten both when its header grows in size as well as when its header shrinks in size. 

## Link to issue
[Vera/ENG-260 Car Stream Safety](https://linear.app/banyan/issue/ENG-260/car-stream-safety)

## Type of change
- [x] This change requires a documentation update

## Test plan (required)
All the myriad test cases that have already been written to ensure CAR stream and `BlockStore` functionality are still present and operational. This functionality is still covered by those test cases.


tomb-common/src/types/blockstore/car/carv1/mod.rs
```diff
@@ -63,23 +62,30 @@ impl CAR {
...
+        // For each offset tallied
+        for block_offset in offsets {
+            // Move to the existing block location
+            rw.seek(SeekFrom::Start(block_offset))?;
+            // Read the block
+            let block = Block::read_bytes(&mut rw)?;
+            // Compute the new offset of the block
+            let new_offset = (block_offset as i64 + data_offset) as u64;
+            // Move to that offset
+            rw.seek(SeekFrom::Start(new_offset))?;
```
amiller68, line 84: I'm pretty sure you can use SeekFrom::Current(data_offset) here

tomb-common/src/types/blockstore/car/carv1/mod.rs
```diff
@@ -63,23 +62,30 @@ impl CAR {
...
+            // Write the block there
+            block.write_bytes(&mut rw)?;
```
amiller68, line 86: What's the rational behind these rewrites? Is it just to move the index back? Why can't we move them all at once?

tomb-common/src/types/blockstore/car/carv1/mod.rs
```diff
@@ -49,10 +49,9 @@ impl CAR {
     }
 
     /// Write out a CARv1 object, assuming the Writer is already seeked to the first byte of the CARv1
```
amiller68, line 51: Comment says Carv1, but indices are related to CarV2. Why are we working index logic into CarV1 stuff?

amiller68 (changes requested):

tomb-common/src/types/blockstore/car/carv1/mod.rs
recursivepaws: We are not presently reading in the Index as it is stored in `CARv2`. To be able to write and read blocks as expected within both our `CARv2`s (which do not process the index) and our `CARv1`s (which do not even store an index) we build a "CARv1 Index" artificially. Doing so requires that we seek through nearly the entire car file, jotting down the CIDs and offsets of those blocks into a HashMap. This is not performant at scale, which is why I am implementing the CARv2 index next thing, but also this code _is_ related to CARv1, it's not the same index you are imagining.

tomb-common/src/types/blockstore/car/carv1/mod.rs
recursivepaws: Moving every Block in the CAR file at once would require that every Block in the CAR file exist in memory- which would break at scale.

tomb-common/src/types/blockstore/car/carv1/mod.rs
recursivepaws: We cannot, because running `Block::read_bytes` moves the stream cursor to the _end_ of the block. If we were to use the current position as the reference frame, we would first need to rewind the entire size of the block, and then add the data_offset, which is equally mathy and slightly more obfuscated than what i'm doing here.

tomb-common/src/types/blockstore/car/carv1/mod.rs
```diff
@@ -63,23 +62,30 @@ impl CAR {
 
         // Keep track of the new index being built
         let mut new_index: HashMap<Cid, u64> = HashMap::new();
+        // Grab all offsets
+        let mut offsets: Vec<u64> = self.index.borrow().clone().map.into_values().collect();
+        // Sort those offsets so the first offsets occur first
+        offsets.sort();
+        // If the header got bigger
+        if data_offset > 0 {
```
amiller68, line 70: Account for data_offset == 0

tomb-common/src/types/blockstore/car/carv1/mod.rs
amiller68: ooop no read moves the cursor my bad

tomb-common/src/types/blockstore/car/carv1/mod.rs
amiller68: nvm, we can get around this later

tomb-common/src/types/blockstore/car/carv1/mod.rs
amiller68: makes sense to me

tomb-common/src/types/blockstore/car/carv1/mod.rs
amiller68: yep yep

amiller68 (approved):
Looks good to me

PR #204, opened 2023-08-17
Vera / ENG-266 Get Tomb Tests Running in CI

recursivepaws:
The primary modifications here are in the `ci.yml` file. 
While I initially believed there might be some error in our code, not able to start from a new clone, it turned out to be the toolchain we were using. In the process of hunting for this bug, though, all unwraps were also replaced with expressive `except` statements.
With the updated nightly toolchain used here, we are now actually testing all of our lints, running wasm tests, running doc tests, and running native tests. 
banyan-api-core related tests are temporarily disabled, for use in a different workflow.


.github/workflows/ci.yml
```diff
@@ -7,72 +7,36 @@ jobs:
...
-      - uses: actions-rs/cargo@v1
-        name: Run all workspace tests
-        with:
-          command: test
-          args: --workspace --lib --bins --tests --benches
-        env:
-          CARGO_INCREMENTAL: 0
-          RUSTFLAGS: "-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Cpanic=abort -Zpanic_abort_tests"
+      - uses: dtolnay/rust-toolchain@nightly
+      - run: RUST_BACKTRACE=1 && cargo test --all-features
```
amiller68: test with default features disabled, (--no-default-features)
remove #[ignore]s from tests in banyan/models

amiller68 (approved):
lgtm

PR #208, opened 2023-08-21
Alex/eng 309 cleaning up and finalizing wasm client (incremental progress)

amiller68:
# Description
Refactors tomb-common to be be more ergonmic / less verbose for downstream libraries

Implements updates to client to reflect close alignment with Spec

Start using Api Client implementation with in  the Wasm Client.


amiller68:
can't merge unitl this upstream PR is merged:
https://github.com/banyancomputer/tomb-crypt/pull/8

tomb-common/src/banyan_api/models/account.rs
```diff
@@ -62,15 +44,26 @@ impl Account {
...
+    pub async fn usage(client: &mut Client) -> Result<u64, ClientError> {
+        let response = client.call(GetTotalUsage).await?;
+        Ok(response.size as u64)
+    }
+
+    /// Get the usage limit for the account associated with the current credentials in the Client
+    pub async fn usage_limit(client: &mut Client) -> Result<u64, ClientError> {
+        let response = client.call(GetUsageLimit).await?;
+        Ok(response.size as u64)
+    }
```
sstelfox, line 58: I would combine these into a single API call, when you need one you'll almost always want the other. I suppose the latter is going to be more infrequent but I don't think they need to be separated.

sstelfox (approved):
There is one change I'd like to see in here but that's for both the API and the client and I don't want to hold up this PR for that API change.

PR #211, opened 2023-08-24
Alex/eng 270 banyan client update

amiller68:
# Description
Eng 270 required I update the client a little bit, to allow pushing metadata without content

### DIFF
--- a/tomb-common/src/banyan_api/models/metadata.rs
+++ b/tomb-common/src/banyan_api/models/metadata.rs
@@ -72,10 +72,10 @@ impl Metadata {
         bucket_id: Uuid,
         root_cid: String,
         metadata_cid: String,
-        data_size: usize,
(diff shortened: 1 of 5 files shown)


tomb-common/src/banyan_api/models/metadata.rs
```diff
@@ -84,24 +84,37 @@ impl Metadata {
...
+                    bucket_id,
+                    root_cid,
+                    metadata_cid,
+                    data_size: 0,
+                    state: response.state,
+                },
+                None,
+            )),
+            Some(_) => Ok((
+                Self {
```
recursivepaws: this Self obj is the same but is defined twice; move outside the match statement for succinctness

recursivepaws (commented):
1 nit otherwise looks good

recursivepaws (approved):

PR #226, opened 2023-09-06
Vera/ENG-225 Public Key Authorization Workflow

recursivepaws:
# Description
- Vastly overhauled CLI to create a more intuitive user experience
- CLI can now interface with `tomb-common` to call `banyan_api` functionality
    - Bucket creation, deletion, information
    - Bucket Key creation, deletion, information
    - Bucket Metadata creation, deletion, information (wip)
    - Bucket Snapshot creation and information (wip)
- BucketConfig get_all / set_all methods replaced with FsMetadata access
    - Public key correlate of the on disk wrapping key is automatically shared with when a bucket is created.

## Link to issue
[Vera/ENG-225 Public Key Authorization Workflow](https://linear.app/banyan/issue/ENG-225/create-public-key-authorization-workflow-in-cli)

## Type of change

## Test plan
Existing CLI and pipeline tests still work. Two tests have been added to the `tomb-common` banyan_api ensure bucket keys can be rejected and approved based on expected api behavior.


tomb-common/src/banyan_api/error.rs
```diff
@@ -1,8 +1,8 @@
 use std::error::Error;
 use std::fmt::{self, Display, Formatter};
-
 use tomb_crypt::prelude::TombCryptError;
 
+/// Errors that can occur in the API Client
```
sstelfox, line 5: Generally I don't find these kinds of "obvious" comments useful but I know there is a lot of them in this repo (and I think there might be a warning turned on for them?). Don't worry about it for now but I wouldn't spend time on them in the future.

tomb-common/src/banyan_api/models/account.rs
```diff
@@ -19,16 +19,16 @@ pub struct Account {
 
 impl Account {
     /// Create a new instance of this model or data structure. Attaches the associated credentials to the client.
-    pub async fn create_fake(_client: &mut Client) -> Result<(Self, EcSignatureKey), ClientError> {
+    pub async fn create_fake(client: &mut Client) -> Result<(Self, EcSignatureKey), ClientError> {
```
sstelfox, line 22: Good catch, I didn't even know you could call variables that had the underscore prefix 😮

tomb-common/src/banyan_api/models/snapshot.rs
```diff
@@ -19,6 +20,18 @@ pub struct Snapshot {
     pub created_at: i64,
 }
 
+impl Display for Snapshot {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
sstelfox, line 24: Hah I do this all the time (including the full path to the module) super useful for quickly moving code around. 👍🏼

tomb/benches/pipeline.rs
```diff
@@ -23,10 +27,10 @@ lazy_static! {
     static ref BENCH_PATH: String = env::var("BENCH_PATH").unwrap_or_else(|_| "target/bench".to_string());
     // Path we will use to hold inputs we build for benchmarking
     static ref INPUT_PATH: String = env::var("INPUT_PATH").unwrap_or_else(|_| "target/bench/input".to_string());
-    // Path we will use to hold packed data we build from inputs for benchmarking
-    static ref PACKED_PATH: String = env::var("PACKED_PATH").unwrap_or_else(|_| "target/bench/packed".to_string());
-    // Path we will use to hold unpacked data we build from packed data for benchmarking
-    static ref UNPACKED_PATH: String = env::var("UNPACKED_PATH").unwrap_or_else(|_| "target/bench/unpacked".to_string());
+    // Path we will use to hold bundleed data we build from inputs for benchmarking
```
sstelfox: There is a typo kind of throughout this file of "bundleed" (with an extra e) instead of bundled. Can't tell if that was intentional or not hahah

tomb/src/main.rs
```diff
@@ -3,6 +3,7 @@
 #![feature(buf_read_has_data_left)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![warn(missing_debug_implementations, missing_docs, rust_2018_idioms)]
+#![deny(private_interfaces)]
```
sstelfox, line 6: Not familiar with this one. Looked it up and seems good to me

tomb/src/pipelines/banyan_api/bucket/keys.rs
```diff
@@ -0,0 +1,102 @@
...
+
+            if let Some(remote_id) = bucket.remote_id {
+                BucketKey::create(remote_id, pem, client)
+                    .await
+                    .map(|key| format!("{}", key))
+                    .map_err(TombError::client_error)
+            } else {
+                Ok("added key to bucket locally".to_string())
+            }
+        }
```
sstelfox: I'm not sure when this creation would be useful and I'm guessing this was what you were asking about before. If this is the approval side of the public key it shouldn't be generating a key, it should be using the public portion of a key already present as a key request on the metadata server.

I don't otherwise see an approve action here or something similar so I'm guessing this is the approval command.

tomb/src/pipelines/banyan_api/bucket/snapshots.rs
```diff
@@ -0,0 +1 @@
+
```
sstelfox: Empty file, this can get cleaned up

tomb/src/utils/mod.rs
```diff
@@ -1,17 +1,17 @@
+/// Utils specific to the Bundle pipeline
+pub mod bundle;
```
sstelfox, line 2: These kinds of renames can really expand the size of these PRs and make the signal-to-noise ratio pretty bad. In the future please keep large-scale renames, or reorganizations in their own PRs away from features and bug fixes.

sstelfox (changes requested):

tomb/benches/pipeline.rs
recursivepaws: this definitely just happened when i replaced 'pack(ed)' with 'bundle(ed)'. will update

tomb/src/main.rs
recursivepaws: yeah, one of the old lints we had became deprecated and this is its successor

tomb/src/pipelines/banyan_api/bucket/keys.rs
recursivepaws: this is basically the computer B scenario, they're connected to a bucket but have their own unapproved key, this 'create bucket key' command really just takes the wrapping key they generate on disk, makes a public key, and sends to server in an unapproved status. might need to be renamed, considering it is only done once per device, and only if you didnt start the bucket.
the manual approval of a bucket key is not an option anymore.

tomb/src/utils/mod.rs
recursivepaws: got it

tomb/src/pipelines/banyan_api/bucket/keys.rs
sstelfox: Ahhh yeah, that's the request access workflow. So maybe adjust the name to something along the lines of request access to be a bit more clear

sstelfox (approved):
Woo 👍🏼

PR #243, opened 2023-09-21
Vera/Mv-Fix

recursivepaws:
- Fixed Mv functionality
- Added Cp & Symlink functionality
- Added get_all_nodes helper method to FsMetadata
- Unified PrivateForests into a single PrivateForest
- Implemented a Split BlockStore for handling extraction cases with large files


tomb-common/src/banyan_api/models/storage_ticket.rs
```diff
@@ -204,16 +204,16 @@ pub mod test {
         let mut fs_metadata = FsMetadata::init(&key)
             .await
             .expect("Failed to create fs metadata");
-        let mkdir_path_segments = vec!["test".to_string(), "path".to_string()];
-        let add_path_segments = vec!["test".to_string(), "path".to_string(), "file".to_string()];
+        let mkdir_path_segments = &vec!["test".to_string(), "path".to_string()];
+        let add_path_segments = &vec!["test".to_string(), "path".to_string(), "file".to_string()];
```
sstelfox: Shouldn't take references of concrete objects to be stored, where they're getting passed around is where you want to be taking references.

tomb-common/src/metadata.rs
```diff
@@ -100,16 +100,9 @@ impl FsMetadata {
...
-        let _metadata_forest_cid =
-            store_forest(&self.metadata_forest, metadata_store, content_store).await?;
-        let content_forest_cid =
-            store_forest(&self.content_forest, content_store, metadata_store).await?;
-        let _content_forest_cid =
-            store_forest(&self.content_forest, content_store, content_store).await?;
-        assert_eq!(metadata_forest_cid, _metadata_forest_cid);
-        assert_eq!(content_forest_cid, _content_forest_cid);
+        let forest_cid = store_forest(&self.forest, metadata_store, metadata_store).await?;
+        let _forest_cid = store_forest(&self.forest, metadata_store, content_store).await?;
```
sstelfox: Underscored variables should never be used/referenced (this is getting used in the assert)

tomb-common/src/metadata.rs
```diff
@@ -649,22 +692,141 @@ mod test {
         let mut fs_metadata =
             _init_save_unlock(wrapping_key, metadata_store, content_store).await?;
 
-        let path = vec!["cat.txt".to_string()];
-        let content = "hello kitty".as_bytes().to_vec();
+        let cat_path = &vec!["cat.txt".to_string()];
```
sstelfox: Here's a &Concrete again

tomb-common/src/metadata.rs
```diff
@@ -423,149 +399,216 @@ impl FsMetadata {
...
+    /// Write data do a specific node
+    pub async fn write(
+        &mut self,
+        path_segments: &[String],
+        metadata_store: &impl RootedBlockStore,
+        content_store: &impl BlockStore,
+        content: Vec<u8>,
+    ) -> Result<()> {
+        let time = Utc::now();
+        let rng = &mut thread_rng();
```
sstelfox: This is the same pattern of taking a reference to an unowned object, this one is mutable so passing it around always ensure the caller gets a mutable version even if they don't need it. It become impossible to tell where mutations occur.

tomb-wasm/src/mount.rs
```diff
@@ -437,7 +437,7 @@ impl WasmMount {
     /// * `Bucket is locked` - If the bucket is locked
     pub async fn ls(&mut self, path_segments: Array) -> TombResult<Array> {
         // Read the array as a Vec<String>
-        let path_segments = path_segments
+        let path_segments = &path_segments
```
sstelfox: Another weird one, and this will actually force an additional copy of the internal strings later on when its joined where ownership of the internal state could just be consumed.

tomb/src/pipelines/extract.rs
```diff
@@ -27,25 +26,54 @@ pub async fn pipeline(
     let config = global.get_bucket_by_specifier(bucket_specifier)?;
     // Load metadata
     let fs = &mut config.unlock_fs(&wrapping_key).await?;
-    let metadata = &config.metadata;
-    let content = &config.content;
+    let metadata_store = &config.metadata;
+    let content_store = &config.content;
```
sstelfox: More inverted references.

sstelfox (changes requested):

tomb-wasm/src/mount.rs
```diff
@@ -609,9 +606,9 @@ impl WasmMount {
         &mut self,
         path_segments: Array,
         _version: Option<String>,
-    ) -> TombResult<ArrayBuffer> {
+    ) -> TombResult<Uint8Array> {
         // Read the array as a Vec<String>
-        let path_segments = path_segments
+        let path_segments = &path_segments
```
sstelfox: This one also shouldn't be a reference.

tomb-wasm/src/mount.rs
```diff
@@ -718,15 +715,15 @@ impl WasmMount {
     /// * `Could not rm` - If the rm fails
     /// * `Could not sync` - If the sync fails
     pub async fn rm(&mut self, path_segments: Array) -> TombResult<()> {
-        let path_segments = path_segments
+        let path_segments = &path_segments
```
sstelfox: And here

sstelfox (changes requested):

sstelfox (approved):

PR #253, opened 2023-10-02
Vera/Auth CLI

recursivepaws:


tomb-common/Cargo.toml
```diff
@@ -30,11 +30,12 @@ sha1 = "^0.10"
 sha2 = "^0.10"
 test-case = "^3"
 thiserror = "^1"
-uuid = { version = "^1", features = ["serde"] }
+uuid = { workspace = true }
```
sstelfox: We do not want to be doing this workspace thing. Using dependencies inside workspaces already share the dependencies when they naturally overlap. It does not enforce the same versioning for dependencies of our dependencies and makes it harder to update individual crates.

tomb-common/Cargo.toml
```diff
@@ -30,11 +30,12 @@ sha1 = "^0.10"
...
 wasm-bindgen = "^0.2"
 unsigned-varint = "^0.7"
 bytes = "1.5.0"
 async-recursion = "1.0.5"
+base64-url = "2.0.0"
```
sstelfox: When adding new dependencies, pin it to the stable semver version (for version > 1.0.0 this is the major version, for version < 1.0.0 it is the minor version. Pinning 2.3.1 would be "^2", pinning 0.34.1 would be "^0.34". This allows flexibility when updating packages and between different dependencies that are pinned to slightly different versions.

tomb-common/src/banyan_api/client.rs
```diff
@@ -165,41 +165,52 @@ impl Client {
...
         &mut self,
         request: T,
     ) -> Result<T::ResponseType, ClientError> {
+        // Determine if this request requires authentication
         let add_authentication = request.requires_authentication();
+        // Create the request builder
         let mut request_builder = request.build_request(&self.remote_core, &self.reqwest_client);
+        // If we need authentication
         if add_authentication {
+            // Obtain the bearer token
```
sstelfox: We generally don't need these detailed comments, the code itself is describing what is going on. Please reserve comments for when there is an assumption, edge case being made or considered that may not be immediately obvious from the code, around code that might seem useless or redundant at first glance that may be accidentally cleaned up in haste later on, or when something more long winded is needed like a discussion on trade offs.

tomb-common/src/banyan_api/requests/core/auth/device_api_key/regwait.rs
```diff
@@ -0,0 +1,61 @@
...
+            .unwrap();
+        // Run a get request
+        client.get(full_url).json(&self)
+    }
+
+    fn requires_authentication(&self) -> bool {
+        // No auth required
+        false
+    }
+}
```
sstelfox: It occurred to me that we have a value we can use to identify these requests better and more simply than a random nonce that should work with the existing registration and can likely be worked around with the existing functionality (though the frontend will still have to play a little bit of a role).

If we switch from a nonce to the fingerprint of the key being registered, we don't need to generate or share an extra value between the client and server. It can be computed directly on either side. The frontend currently puts the key directly in the database and would need to make an extra HTTP call to the core service (authenticated using the browser's API device key credentials) so that a signal can be sent to this waiting connection... But that is a much smaller lift than moving everything over and gets this working much quicker.

tomb-wasm/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde = { workspace = true }
 tomb-crypt = { workspace = true }
 wnfs = { workspace = true }
 blake3 = { workspace = true }
+uuid = { workspace = true }
```
sstelfox: Only enforcing this on new dependencies, we'll want to clean up the old ones eventually but yeah as I said elsewhere, for new dependencies please don't add them workspace wide.

tomb/Cargo.toml
```diff
@@ -18,7 +18,7 @@ serde_json = { workspace = true }
 serde = { workspace = true }
 tokio = { workspace = true }
 tomb-crypt = { workspace = true }
-uuid = "1.4.1"
+uuid = { workspace = true }
```
sstelfox: See other comments on workspace deps

tomb/Cargo.toml
```diff
@@ -31,6 +31,8 @@ env_logger = "^0.10"
 async-trait = "^0.1"
 hex = "^0.4"
 tomb-common = { path = "../tomb-common", version = "^1", features = ["banyan-api"] }
+open = "5.0.0"
+base64-url = "2.0.0"
```
sstelfox: See other comment on version pinning.

I also believe we have base64 already present as a dependency which supports web safe Base64 encoding as well and should likely be used instead of this.

tomb/src/pipelines/banyan_api/auth.rs
```diff
@@ -15,15 +16,107 @@ pub async fn pipeline(command: AuthSubCommand) -> Result<String> {
...
+                .strip_suffix("-----END PUBLIC KEY-----")
+                .expect("unable to strip PEM suffix")
+                .to_string();
+
+            // Represent the weird b64 characters with ones that are url-valid
+            let encoded_public_key = stripped_public_key
+                .replace('+', "-")
+                .replace('/', "_")
+                .replace('=', ".")
+                .to_string();
```
sstelfox: This looks like its just getting the DER bytes from the public key. I think there is a helper method for that already

tomb/src/pipelines/banyan_api/auth.rs
```diff
@@ -15,15 +16,107 @@ pub async fn pipeline(command: AuthSubCommand) -> Result<String> {
...
+
+            // Create a new nonce to identify this registration
+            let nonce = Uuid::new_v4();
```
sstelfox: I mentioned this elsewhere but after reading through this PR I think we can use the fingerprint of the public key as a shared value here instead of a randomly generated nonce. It'll be consistent and not have to be exchanged between the endpoints.

tomb/src/pipelines/banyan_api/auth.rs
```diff
@@ -15,15 +16,107 @@ pub async fn pipeline(command: AuthSubCommand) -> Result<String> {
...
+            let mut client_1 = client.clone();
+            // Create a join handle for later use, starting the call immediately
+            let join_handle: JoinHandle<Result<RegwaitResponse, ClientError>> =
+                tokio::task::spawn(async move {
+                    // Build the request
+                    client_1.call(start_regwait).await
+                });
+
+            // Should be this in prod TODO
+            // https://alpha.data.banyan.computer/
```
sstelfox: Yeah this will need to be configurable somehow. We're likely going to want it as part of the CLI args with a default of whatever our current "production environment" is.

sstelfox (changes requested):

tomb-common/src/banyan_api/requests/core/auth/device_api_key/regwait.rs
recursivepaws: noted; will implement

Cargo.toml
```diff
@@ -49,7 +49,8 @@ reqwest = { version = "^0.11", features = ["blocking", "multipart", "json", "str
 serde_json = { version = "^1", features = ["std"]}
 serde = {version= "^1", features = ["derive"]}
 tokio = { version = "^1", features = ["full", "io-util", "fs"]}
-tomb-crypt = { git = "https://github.com/banyancomputer/tomb-crypt", branch = "main", version = "^0.1" }
+#tomb-crypt = { git = "https://github.com/banyancomputer/tomb-crypt", branch = "alex/rust-crypto", version = "^0.1" }
+tomb-crypt = { path = "../tomb-crypt-new/tomb-crypt" }
```
sstelfox: We probably want this pointing at a specific ref of the remote github repo until the relevant PR gets merged. Relative paths are hard to reproduce on different machines and easy to get out of sync.

tomb-common/src/banyan_api/client.rs
```diff
@@ -165,41 +165,50 @@ impl Client {
...
+    /// Call a method that implements ApiRequest on the core server
     pub async fn call<T: ApiRequest>(
         &mut self,
         request: T,
     ) -> Result<T::ResponseType, ClientError> {
+        // Determine if this request requires authentication
         let add_authentication = request.requires_authentication();
         let mut request_builder = request.build_request(&self.remote_core, &self.reqwest_client);
         if add_authentication {
+            // Obtain the bearer token
```
sstelfox: (I won't comment on this again because I don't really mind)

These little micro comments don't really add any value so I wouldn't spend extra time on them and would generally prefer comments describing things that can be easily be inferred such as this called to bearer_token() obtaining a bearer token be omitted. They do have an ongoing cost to maintain as code changes.

Generally what I prefer doing when there is a chunk of code that isn't obvious what it's behavior is doing or for, is to give it a name by turning the relevant code into a function. It has the added benefit of allowing you to scan a much smaller amount of code to figure out what needs to be done. Functions do have a tiny amount of overhead but that performance hit is so down in the noise its always worth the trade off for future developer productivity.

I'm definitely guilty of not breaking functions up enough to be as readable as I'd prefer and its a good habit. Something to keep in mind but you don't need to do anything here for this review.

tomb-common/src/banyan_api/models/account.rs
```diff
@@ -71,6 +71,10 @@ pub mod test {
         client
     }
 
+    pub async fn unauthenticated_client() -> Client {
+        Client::new("http://localhost:3001", "http://localhost:3002").unwrap()
```
sstelfox, line 74: We don't want to be hardcoding these values if they're used in the general client, if this function is only used for testing it should be kept out of prod code with a cfg attribute.

tomb-common/src/banyan_api/requests/core/auth/device_api_key/regwait/end.rs
```diff
@@ -0,0 +1,70 @@
...
+impl ApiRequest for EndRegwait {
+    type ResponseType = EndRegwaitResponse;
+    type ErrorType = EndRegwaitError;
+
+    fn build_request(
+        self,
+        base_url: &reqwest::Url,
+        client: &reqwest::Client,
+    ) -> reqwest::RequestBuilder {
+        println!("base_url: {:?}", base_url);
```
sstelfox: Diagnostic left in

tomb/src/pipelines/banyan_api/auth.rs
```diff
@@ -54,3 +74,114 @@ pub async fn pipeline(command: AuthSubCommand) -> Result<String> {
...
+                .map(|v| v.map_err(anyhow::Error::new))??;
+
+            // Update the client's credentials
+            Ok(Credentials {
+                account_id: start_response.account_id,
+                signing_key: private_device_key,
+            })
+        }
+        Err(_) => Err(anyhow::anyhow!("end_regwait failed")),
+    }
```
sstelfox: This shouldn't be making a call to the endregwait, that should be getting made by the frontend or tomb-wasm after the key has been registered in the database. This just needs to wait for the start call to complete (which will give you the account ID) or for it to timeout/be rejected (which will return an error).

sstelfox (changes requested):

sstelfox:
Also looks like there are some merge conflict with the main branch that need to get resolved (probably just me locking tomb-crypt to the main version which you can update in this branch to point to the correct revision).

tomb-wasm/src/lib.rs
```diff
@@ -61,7 +63,7 @@ impl TombWasm {
     /// A new TombWasm instance
     ///
     /// Don't call it from multiple threads in parallel!
-    pub fn new(
+    pub async fn new(
```
sstelfox, line 66: This change is a break in the API that javascript is currently expecting. Is there an equivalent change on the other side?

tomb/src/pipelines/banyan_api/auth.rs
```diff
@@ -15,15 +21,29 @@ pub async fn pipeline(command: AuthSubCommand) -> Result<String> {
...
             let api_key = EcSignatureKey::generate().await?;
             let public_api_key = api_key.public_key()?;
             let public_api_key_pem = String::from_utf8(public_api_key.export().await?)?;
             // Associate the key material with the backend
             let response: CreateAccountResponse = client
-                .call(CreateAccount {
+                .call(CreateFakeAccount {
                     device_api_key_pem: public_api_key_pem,
                 })
                 .await?;
```
sstelfox, line 49: I don't think its supposed to be using the Fake account creation outside of tests right?

sstelfox (changes requested):
Couple of questions left outstanding before a merge 👍🏼 looks good though

tomb-wasm/src/lib.rs
recursivepaws: there is an equivalent change in the core repo yes

tomb/src/pipelines/banyan_api/auth.rs
recursivepaws: true, this branch of the match statement only exists if you have the "fake" feature on- which is an easy way to ensure that you can still interact with the fake registration auth stuff through the CLI if you want to, but it doesn't work or present itself as an option in clap unless the feature is set

sstelfox (approved):

PR #259, opened 2023-10-20
Vera/ENG-308 Data Restoration

recursivepaws:
# Description
### Huge refactor of CLI structure:
- New `OmniBucket` class now holds two optional fields, one being a `LocalBucket` and the other being a `RemoteBucket`. 
    - This new struct is used in CLI output as well as important operations
- Bucket Syncing now uses a `SyncState` enum to assess and articulate the current status of a local and remote bucket in comparison to each other. 
- Metadata pushing and pulling is now integrated into this syncing in a context-aware way rather than the previous method which required users to manually push metadata themselves.
- Data reconstruction appears almost entirely functional- there is something wrong with `PrivateForest` reconstruction that although reproducible is hard to understand as it is not an issue in WASM presently. Currently syncing with Phillip in search of a fix. 
- Now using the latest version of `tomb-crypt`, making this repo home to the necessary WebCrypto helper methods.

## Link to issue
https://linear.app/banyan/issue/ENG-308/data-restoration

## Type of change

- [ ] This change requires a documentation update

## Test plan
Most of this code is integration based and should not be tested through cargo tests but rather through actually interacting with the tool.


sstelfox (approved):

sstelfox:
Looks good 👍🏼

PR #266, opened 2023-10-24
Alex/mv fix

amiller68:
# Description
Fixes mv / adds a few tests / marks a couple of long ones as ignored


tomb-common/Cargo.toml
```diff
@@ -18,6 +18,7 @@ serde = { workspace = true }
 tomb-crypt = { workspace = true } 
 wnfs = { workspace = true }
 blake3 = { workspace = true }
+cid = { workspace = true }
```
sstelfox: New dependencies shouldn't be using workspace versions.

sstelfox: (these all need to go away but not in this PR)

sstelfox (changes requested):

sstelfox (approved):

PR #267, opened 2023-10-26
Vera/ENG-308 Data Restoration

recursivepaws:
This includes the already approved changes from [this PR](https://github.com/banyancomputer/tomb/pull/265). 

as such most of the changes in this PR are block tracking and the new rustfmt. full reconstruction functionality will not be available until claudia/screaming is merged on tomb and refactor-compat is merged on core alongside it.


tomb-common/src/metadata.rs
```diff
@@ -324,7 +325,10 @@ impl FsMetadata {
             .get_node(path_segments, true, &self.forest, metadata_store)
             .await;
 
-        if let Ok(node) = result && node.is_some() {}
+        if let Ok(node) = result
+            && node.is_some()
+        {
+        }
         // If there was an error searching for the Node or
         else {
```
sstelfox, line 333: This is an odd ball... I think the error portion of this should be an early return... Right now it looks like you're attempting to read from the metadata store and if an error occurs fall back on an instance of the split store that wraps both the metadata store and the content store... Why not just use that directly the first time and return early if it returns an error?

tomb/src/pipelines/reconstruct.rs
```diff
@@ -68,7 +57,9 @@ pub async fn pipeline(
                     symlink(origin, built_path)?;
                 } else {
                     // If the parent does not yet exist
-                    if let Some(parent) = built_path.parent() && !parent.exists() {
+                    if let Some(parent) = built_path.parent()
+                        && !parent.exists()
+                    {
```
sstelfox, line 62: Same as before... This is a really odd pattern and probably needs to be broken up into an early return

tomb/src/types/config/bucket/omni.rs
```diff
@@ -64,7 +64,9 @@ impl OmniBucket {
             new_object.local = Some(bucket);
         }
         // If we found one
-        if let Some(remote_id) = remote_id && let Ok(bucket) = RemoteBucket::read(client, remote_id).await {
+        if let Some(remote_id) = remote_id
+            && let Ok(bucket) = RemoteBucket::read(client, remote_id).await
+        {
```
sstelfox, line 69: This is kind of even weirder than the others, definitely split this up.

tomb/src/types/config/bucket/omni.rs
```diff
@@ -140,19 +132,20 @@ impl OmniBucket {
...
             remote: None,
             sync_state: None,
         };
         // If this bucket already exists both locally and remotely
-        if let Some(bucket) = global.get_bucket(origin) &&
-            let Some(remote_id) = bucket.remote_id &&
-            RemoteBucket::read(client, remote_id).await.is_ok() {
+        if let Some(bucket) = global.get_bucket(origin)
+            && let Some(remote_id) = bucket.remote_id
+            && RemoteBucket::read(client, remote_id).await.is_ok()
```
sstelfox, line 143: Same pattern

tomb/src/types/config/bucket/omni.rs
```diff
@@ -204,8 +195,9 @@ impl OmniBucket {
         global: &mut GlobalConfig,
         client: &mut Client,
     ) -> Result<String, TombError> {
-        let local_deletion = if let Some(local) = &self.local &&
-        prompt_for_bool("are you sure you want to delete this Bucket locally?") {
+        let local_deletion = if let Some(local) = &self.local
+            && prompt_for_bool("are you sure you want to delete this Bucket locally?")
+        {
```
sstelfox, line 200: And here (looks like there are a bunch more, I'm going to stop commenting on the individual instances)

tomb/src/types/config/bucket/omni.rs
```diff
@@ -286,19 +280,20 @@ impl Display for OmniBucket {
...
                 remote.r#type,
                 remote.storage_class
             );
-        }
-        else if let Some(local) = &self.local {
+        } else if let Some(local) = &self.local {
             info = format!("{info}\n{}", local);
-        }
-        else if let Some(remote) = &self.remote {
+        } else if let Some(remote) = &self.remote {
```
sstelfox, line 296: This is slightly different from before but is much harder to follow with the extra complexity in the else if statements.

sstelfox (changes requested):

sstelfox (changes requested):
Letting the `if let` chains I commented on be for now. We'll want to address those when dealing with our tech debt.

sstelfox (approved):

PR #269, opened 2023-10-29
Vera/Storage Tickets

recursivepaws:
- Removed duplicate code in favor of architecture dependent type aliases
- Moved disk-based blockstores out of tomb-common and into tomb
    - Prevents us from having to architecture guard those modules for fear of breaking wasm
    - Allows us to write the trait implementations for them in tomb cli, where they are being used
- Removed `upload_content` as a function on `StorageTicket`, instead replacing it with a trait that implements uploading.
    - Removes the need to use duplicate content hashing and length getting code every time we upload
    - Allows us to perform an upload even when a storage ticket is not provided but a host is (cases where new grants are not needed and user already has available capacity on server)


tomb-common/Cargo.toml
```diff
@@ -50,5 +50,8 @@ hex-literal = "^0.4"
 crate-type = ["cdylib", "rlib"]
 path = "src/lib.rs"
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+tokio = { workspace = true }
+
```
sstelfox: We don't want to be using workspace specified dependencies, it's tech-debt that we already have so many. The lockfile is already shared so as long as the major semver is the same we get the useful portion of this feature automatically without tightly binding together the workspace deps in hard to untangle ways.

tomb-wasm/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 //! This crate contains modules which are compiled to WASM
+#![feature(let_chains)]
```
sstelfox: Please don't use these. Part of the tech debt I'm going to have people working on while I'm gone is getting these crates compiling on stable rust and removing these non-standard-for-a-very-good-reason extensions to the syntax.

tomb-common/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![feature(let_chains)]
 #![feature(file_create_new)]
 #![feature(duration_constants)]
+#![feature(type_alias_impl_trait)]
```
sstelfox: Please don't use these. Part of the tech debt I'm going to have people working on while I'm gone is getting these crates compiling on stable rust and removing these non-standard-for-a-very-good-reason extensions to the syntax.

sstelfox (changes requested):
I know this is only a draft PR but I was already reviewing opened PRs and saw a few things already worth calling out.

sstelfox (approved):
Looks good 👍🏼

PR #273, opened 2023-10-30
feat: use pems in wasm client api

amiller68:
# Description
Makes the tomb-wasm client use PEM formatted strings in place of CrytpoKeys or CryptoKeyPairs


amiller68:
Not sure if this is the proper time to merge. this would block incorporating newer changes into the banyan-core frontend until the passing of keys back and form between js and wasm is refactored

PR #280, opened 2023-11-06
Alex/eng 412 update api to use user id in place of account

amiller68:
# Description
static site integration already has an updated build.
i don't want to merge this till API updates from that are in main on banyan-core
creating a draft to track this


recursivepaws (approved):
looks good tests good :D

PR #290, opened 2023-11-20
feat: rename buckets

amiller68:
# Description
Implements API request for updating Bucket through a PUT request. Right now the core server only accepts updates to `name`

Implements methods and tests for two possible workflows for renaming a bucket in wasm module
- through the highest level api client
- through the mount client


sstelfox (approved):
Pretty straight forward 👍🏼

PR #291, opened 2023-11-21
Vera/ENG-355 File Sharing

recursivepaws:
# Description
Add the ability to share a Snapshot of a File from an `FsMetadata` instance, or to reconstruct one if you have the appropriate forest, blockstore, and `SharedFile` struct. 

## Link to issue
https://linear.app/banyan/issue/ENG-355/file-sharing

## Type of change

## Test plan (required)
There is a simple test implemented in `FsMetadata` that checks for the viability of this functionality.


src/wasm/compat/mount.rs
```diff
@@ -850,6 +851,47 @@ impl WasmMount {
         Ok(())
     }
 
+    /// Share a file snapshot
+    #[wasm_bindgen(js_name = shareFile)]
+    pub async fn share_file(&mut self, path_segments: Array) -> TombResult<String> {
```
recursivepaws, line 854: should this just return a `WasmSharedFile`? I'm not sure if its better to just spit out the b64 data right here or somewhere else, like leaving that to the frontend

src/filesystem/sharing/shared_file.rs
```diff
@@ -0,0 +1,66 @@
+use base64::{engine::general_purpose::URL_SAFE, Engine as _};
+use serde::{Deserialize, Serialize};
+use wnfs::{common::dagcbor, libipld::Cid, private::share::SharePayload};
+
+use super::SharingError;
+
+#[derive(Debug, Clone)]
+pub struct SharedFile {
```
recursivepaws, line 7: This includes the content_cid, snapshot_key, and label HashOutput internal to the SharePayload. 
The forest_cid references the serialized IPLD version of the forest. If we can't grab that Block from somewhere, then we presumably cant grab any of the blocks it references either. so this is my idea of a more space-efficient export than just actually exporting the forest with it. 
that said, there are no block locations in this struct. should there be?

src/filesystem/sharing/shared_file.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub mime_type: String,
+    pub size: u64,
+}
+
+impl Serialize for SharedFile {
+    fn serialize<S>(&self, serializer: S) -> std::prelude::v1::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let payload_bytes = dagcbor::encode(&self.payload).expect("failed to serialize payload");
```
recursivepaws: of course, the inbuilt WNFS serializer that runs via a `#[derive(..)]` is asymetric and cant reconstruct. so i'm getting around that by implementing a custom serialization on this type.

src/wasm/compat/mod.rs
```diff
@@ -424,4 +428,15 @@ impl TombWasm {
         // Ok
         Ok(mount)
     }
+
+    #[wasm_bindgen(js_name = receiveFile)]
+    pub async fn receive_file(&mut self, shared_file: WasmSharedFile) -> TombResult<Uint8Array> {
```
recursivepaws: receiving files shouldn't require auth- is taking a &mut self here appropriate? it seems like having this function here as opposed to mount.rs is the better choice, and it seems like we will need a `Client` to reconstruct via the remote blockstore.

src/wasm/compat/mount.rs
sstelfox: The frontend is not going to be re-creating the files themselves, it needs the data that it can put into a URL for the web service to use. All of the data required to locate and decrypt the file needs to be contained within this string.

src/wasm/compat/mod.rs
sstelfox: I'm not actually sure what this is here for... The WASM world doesn't need to do this reconstruction, that is going to happen in the web service.

src/filesystem/metadata.rs
```diff
@@ -333,10 +404,8 @@ impl FsMetadata {
             .get_node(path_segments, true, &self.forest, metadata_store)
             .await;
 
-        if let Ok(Some(_)) = result {
-        }
-        // If there was an error searching for the Node or
-        else {
+        // If there was an error getting the node
+        if result.is_err() || result.unwrap().is_none() {
```
sstelfox: We don't want to be adding more unwraps() or expects() wherever possible and this looks like a good place to avoid a potential panic.

src/filesystem/metadata.rs
```diff
@@ -259,6 +261,75 @@ impl FsMetadata {
...
+        // Extract relevant bits of metadata
+        let metadata = file.get_metadata();
+        let mime_type = match metadata.0.get("mime_type") {
+            Some(Ipld::String(mime_type)) => mime_type.to_owned(),
+            _ => "NA".to_owned(),
+        };
+        let size = match metadata.0.get("size") {
+            Some(Ipld::Integer(size)) => *size as u64,
+            _ => 0,
+        };
```
sstelfox: We should probably keep the Option types here especially with a JSON encoding. The serialized type can use `skip_serializing_if = Option::is_none` as a serde attribute to prevent the field from being encoded. We really want to keep this string small.

src/filesystem/sharing/shared_file.rs
sstelfox: It's kind of a trade off whether blocks should be included... Knowing them ahead of time means we can proactively pull them all at once in parallel, this is a slower way as we have to retrieve the blocks associated with the forest, decrypt it then retrieve the blocks one at a time (there is some possibility for parallelism here once the forest block is decoded).

That being said files can have A LOT of blocks and I don't think we could fit all of those CIDs into a URL... So yeah I think we'll have to stick with the slow path for now (no changes needed).

sstelfox (changes requested):

sstelfox (approved):
Looks good 👍🏼

PR #292, opened 2023-11-21
Vera/ENG-402 banyan-cli Stable Compilation

recursivepaws:
# Description
Removed all features and got library compiling on stable. Updated CI to test for the same.

## Link to issue
https://linear.app/banyan/issue/ENG-402/banyan-cli-stable-compilation

## Type of change

### DIFF
--- a/src/blockstore/carv2_disk.rs
+++ b/src/blockstore/carv2_disk.rs
@@ -36,26 +36,25 @@ impl CarV2DiskBlockStore {
             }
 
             // If the file is already a valid CarV2v2
-            if let Ok(mut file) = File::open(path)
-                && let Ok(car) = CarV2::read_bytes(&mut file)
-            {
-                Ok(Self {
-                    path: path.to_path_buf(),
-                    car,
-                })
+            if let Ok(mut file) = File::open(path) {
+                if let Ok(car) = CarV2::read_bytes(&mut file) {
+                    return Ok(Self {
+                        path: path.to_path_buf(),
+                        car,
+                    });
+                }
             }
+
             // If we need to create the CarV2v2 file from scratch
-            else {
-                // Grab read and write
-                let mut rw = get_read_write(path)?;
-                // Create new
-                let store = CarV2DiskBlockStore {
-                    path: path.to_path_buf(),
-                    car: CarV2::new(&mut rw)?,
-                };
-                // Return Ok
-                Ok(store)
-            }
+            // Grab read and write
+            let mut rw = get_read_write(path)?;
+            // Create new
+            let store = CarV2DiskBlockStore {
+                path: path.to_path_buf(),
+                car: CarV2::new(&mut rw)?,
+            };
+            // Return Ok
+            Ok(store)
         }
     }
 
--- a/src/blockstore/multi_carv2_disk.rs
+++ b/src/blockstore/multi_carv2_disk.rs
@@ -37,18 +37,18 @@ impl MultiCarV2DiskBlockStore {
 
             let mut deltas = Vec::new();
             // For each child in the directory
-            for dir_entry in fs::read_dir(dir)? {
+            for dir_entry in fs::read_dir(dir)?.flatten() {
                 // If the dir entry is valid, the file is a .car, and a MultiCarV2DiskBlockStore can be read from it
-                if let Ok(entry) = dir_entry
-                    && entry
-                        .file_name()
-                        .to_str()
-                        .expect("no file name str")
-                        .contains(".car")
-                    && let Ok(car) = CarV2DiskBlockStore::new(&entry.path())
+                if dir_entry
+                    .file_name()
+                    .to_str()
+                    .expect("no file name str")
+                    .contains(".car")
                 {
-                    // Push this to the vec
-                    deltas.push(car);
+                    if let Ok(car) = CarV2DiskBlockStore::new(&dir_entry.path()) {
+                        // Push this to the vec
+                        deltas.push(car);
+                    }
                 }
             }
 
@@ -150,7 +150,6 @@ impl UploadContent for MultiCarV2DiskBlockStore {
         Ok(hasher.finalize().to_string())
     }
 
-    #[allow(refining_impl_trait)]
     async fn get_body(&self) -> Result<ContentType, Self::UploadError> {
         Ok(tokio::fs::File::open(&self.get_delta()?.path).await?.into())
     }
--- a/src/car/v1/mod.rs
+++ b/src/car/v1/mod.rs
@@ -40,14 +40,15 @@ impl CarV1 {
         // End of the header
         let header_end = r.stream_position()?;
         // If we're in a CARv2
-        if let Some(index_offset) = index_offset
-            && r.seek(SeekFrom::Start(index_offset)).is_ok()
-            && let Ok(index) = <Index<Bucket>>::read_bytes(&mut r)
-        {
-            return Ok(Self {
-                header,
-                index: RefCell::new(index),
-            });
+        if let Some(index_offset) = index_offset {
+            if r.seek(SeekFrom::Start(index_offset)).is_ok() {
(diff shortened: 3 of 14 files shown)


amiller68 (approved):

amiller68:
looks good 👍

src/blockstore/carv2_disk.rs
```diff
@@ -36,26 +36,25 @@ impl CarV2DiskBlockStore {
...
-                    car,
-                })
+            if let Ok(mut file) = File::open(path) {
+                if let Ok(car) = CarV2::read_bytes(&mut file) {
+                    return Ok(Self {
+                        path: path.to_path_buf(),
+                        car,
+                    });
+                }
             }
```
sstelfox, line 46: Failing to open a file, and failing to read bytes seem like they should both be handled errors. It looks like there is a fallback path for the case but the logic is hard to follow here. If it fails to open the path its falling back on a separate function `get_read_write` to attempt a different way to open the file? Why would one fail and the other succeed? Can you distinguish between the failures when the caller handles the error? Is this something the caller should know to handle?

Why are there two separate ways here for getting a CarV2 object (read_bytes vs new)?

src/blockstore/multi_carv2_disk.rs
```diff
@@ -37,18 +37,18 @@ impl MultiCarV2DiskBlockStore {
...
-                        .file_name()
-                        .to_str()
-                        .expect("no file name str")
-                        .contains(".car")
-                    && let Ok(car) = CarV2DiskBlockStore::new(&entry.path())
+                if dir_entry
+                    .file_name()
+                    .to_str()
+                    .expect("no file name str")
+                    .contains(".car")
```
sstelfox, line 46: We probably want an ends_with here. We can't handle a `.car.gz` for example.

src/blockstore/multi_carv2_disk.rs
```diff
@@ -37,18 +37,18 @@ impl MultiCarV2DiskBlockStore {
 
             let mut deltas = Vec::new();
             // For each child in the directory
-            for dir_entry in fs::read_dir(dir)? {
+            for dir_entry in fs::read_dir(dir)?.flatten() {
```
sstelfox, line 40: I'm not confident flatten() is the right move here but its not terribly important and I want to get these comments out.

src/car/v1/mod.rs
```diff
@@ -40,14 +40,15 @@ impl CarV1 {
...
-            });
+        if let Some(index_offset) = index_offset {
+            if r.seek(SeekFrom::Start(index_offset)).is_ok() {
+                if let Ok(index) = <Index<Bucket>>::read_bytes(&mut r) {
+                    return Ok(Self {
+                        header,
+                        index: RefCell::new(index),
+                    });
+                }
+            }
```
sstelfox, line 51: This has the same problem as the disk store, there are errors here that should probably be handled. The documentation claims the car file should already be seeked to the correct location so I'm not sure why the offset is being provided here or any seeking is occurring...

src/car/v2/index/mod.rs
```diff
@@ -107,15 +107,17 @@ impl Index<Bucket> {
...
+            if let Ok((varint, cid)) = Block::start_read(&mut *r) {
+                // Log where we found this block
+                new_index.insert_offset(&cid, block_offset);
+                // Skip the rest of the block
+                r.seek(SeekFrom::Current(
+                    varint as i64 - cid.to_bytes().len() as i64,
+                ))?;
+            } else {
+                break;
+            }
```
sstelfox, line 120: The failure case should be an early return with the happy path being the final case.

sstelfox (commented):
Some changes, mostly not related directly to the stable but we're already touching these files.

src/blockstore/carv2_disk.rs
recursivepaws: calling `new` on the Carv2diskblockstore will check if there is already a valid CAR file at the path given. only if those two `if let`s succeed is that the case. 
otherwise, the `new` method will attempt to create a new file on disk and serialize a new, default CARv2 to it. 
so while i can see the logic here being confusing, the order of operations here still make sense to me

sstelfox: Creation should be an intentional choice under known conditions, not a byproduct of a failed read...

src/filesystem/sharing/mapper.rs
```diff
@@ -115,19 +115,23 @@ impl EncRefMapper {
...
-                }
+                };
+                let Some(Ipld::Bytes(public_key)) = sub_map.get(PUBLIC_KEY_LABEL) else {
+                    return Err(SharingError::unauthorized());
+                };
+                let Some(Ipld::String(encrypted_private_ref)) =
+                    sub_map.get(ENCRYPTED_PRIVATE_REF_LABEL)
+                else {
+                    return Err(SharingError::unauthorized());
+                };
```
sstelfox, line 128: These else's are a very weird pattern especially combined with the destructuring. The else can be removed with a `ok_or` that can return the error itself with the `?` operator which removes the Option type cleanly. You'll want to provide the SharingError variant directly instead of using a construction method (which shouldn't really be used for any error variants that don't take an argument).

src/native/configuration/globalconfig.rs
```diff
@@ -74,15 +74,12 @@ impl GlobalConfig {
...
-            Ok(Credentials {
+        match (self.api_key().await, self.remote_user_id) {
+            (Ok(signing_key), Some(user_id)) => Ok(Credentials {
                 signing_key,
                 user_id,
-            })
-        } else {
-            Err(NativeError::missing_credentials())
+            }),
+            _ => Err(NativeError::missing_credentials()),
```
sstelfox, line 82: The tuple construction should be delayed here and the two failure conditions should be teased apart into independent errors. Errors should represent a unique failure point that can easily be teased apart.

src/native/configuration/globalconfig.rs
```diff
@@ -136,15 +133,15 @@ impl GlobalConfig {
...
-            Ok(config)
+        if let Ok(file) = get_read(&config_path()) {
+            if let Ok(config) = serde_json::from_reader(file) {
+                return Ok(config);
+            }
         }
+
+        let config = Self::create().await?;
+        config.to_disk()?;
+        Ok(config)
```
sstelfox, line 144: These hidden failure modes are impossible to reason about without diving into the code. This is one of those patterns that needs to go

src/native/sync/omni.rs
```diff
@@ -140,12 +140,13 @@ impl OmniBucket {
...
-        {
-            // Prevent the user from re-creating it
-            return Err(NativeError::unique_error());
+        if let Some(bucket) = global.get_bucket(origin) {
+            if let Some(remote_id) = bucket.remote_id {
+                if RemoteBucket::read(client, remote_id).await.is_ok() {
+                    // Prevent the user from re-creating it
+                    return Err(NativeError::unique_error());
+                }
+            }
```
sstelfox, line 149: Is this invisibly updating the internal global state? Those are scary side-effects that are hard to reason about and violates the expectations of anyone using the library if they're not clearly spelled out. The form of this tells me that I _shouldn't expect side effects_ but the useful values seem to be getting discarded.

src/native/sync/omni.rs
```diff
@@ -191,24 +192,24 @@ impl OmniBucket {
...
         }
 
-        if let Ok(remote) = self.get_remote()
-            && remote_deletion
-        {
-            remote_deletion = RemoteBucket::delete_by_id(client, remote.id).await.is_ok();
+        if remote_deletion {
+            if let Ok(remote) = self.get_remote() {
+                remote_deletion = RemoteBucket::delete_by_id(client, remote.id).await.is_ok();
+            }
```
sstelfox, line 212: This looks like another silent failure case. The remote not being present should probably be an error here. If the user of the library is expecting to remotely make a change, then they expect it to exist. Assumptions made by user and developers consuming a library assumptions should be explicitly managed through communication.

PR #299, opened 2023-11-29
Vera/Stable [Sam's Feedback]

recursivepaws:
# Description
[This PR](https://github.com/banyancomputer/banyan-cli/pull/292) was merged prematurely. This PR attempts to implement the suggestions outlined there.

### DIFF
--- a/src/blockstore/carv2_disk.rs
+++ b/src/blockstore/carv2_disk.rs
@@ -35,7 +35,7 @@ impl CarV2DiskBlockStore {
                 File::create(path)?;
             }
 
-            // If the file is already a valid CarV2v2
+            // If the file is already a valid CARv2
             if let Ok(mut file) = File::open(path) {
                 if let Ok(car) = CarV2::read_bytes(&mut file) {
                     return Ok(Self {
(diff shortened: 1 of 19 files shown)


src/utils/cast.rs
```diff
@@ -0,0 +1,10 @@
+#[macro_export]
```
recursivepaws, line 1: this kind of macro appears useful to me. you commented that i should `ok_or` when deconstructing those variables, but unless i'm mistaken there isnt a nice inbuilt way to do that. given that we unwrap enums this way pretty frequently in the project, having a shorthand for it could save us some time. but open to suggestions either way

src/blockstore/carv2_disk.rs
```diff
@@ -24,38 +24,36 @@ pub struct CarV2DiskBlockStore {
...
+            return Err(BlockStoreError::missing_file(path));
         }
+
+        // If the file is already a valid CARv2
+        let mut file = File::open(path)?;
+        let car = CarV2::read_bytes(&mut file)?;
+        Ok(Self {
+            path: path.to_path_buf(),
+            car,
+        })
```
sstelfox, line 56: Wooo! Yeah much better

src/utils/cast.rs
sstelfox: Ah I missed that this was deconstructing the enum type as well. Yeah this is a fine helper.

sstelfox (approved):

PR #307, opened 2023-12-04
Vera/ENG-428 Remove Staging URL

recursivepaws:
# Description
The `Client` struct no longer takes a staging url, and likewise the `Endpoints` struct serialized in configs no longer serializes one. Instead we rely on storage tickets and block location requests exclusively. 

## Link to issue
https://linear.app/banyan/issue/ENG-428/remove-staging-url

## Type of change


src/native/configuration/endpoints.rs
```diff
@@ -16,13 +14,11 @@ impl Default for Endpoints {
         if option_env!("DEV_ENDPOINTS").is_some() {
             Self {
                 core: "http://127.0.0.1:3001".to_string(),
-                data: "http://127.0.0.1:3002".to_string(),
-                frontend: "http://127.0.0.1:3000".to_string(),
+                frontend: "http://127.0.0.1:3001".to_string(),
```
sstelfox: We can get rid of the frontend as well, this will always be the core service URL as well (Its available at both the api. domain and the beta/alpha domain but that's all combined into one now).

src/native/operations/mod.rs
```diff
@@ -46,7 +46,7 @@ mod test {
     async fn prepare_pipeline(origin: &Path) -> Result<String, NativeError> {
         let mut global = GlobalConfig::from_disk().await?;
         let wrapping_key = global.wrapping_key().await?;
-        let mut client = Client::new("http://google.com", "http://google.com")?;
+        let mut client = Client::new("http://google.com")?;
```
sstelfox: We probably shouldn't be sending random traffic to Google even in tests...

sstelfox (changes requested):

sstelfox (approved):
Looks good!

PR #309, opened 2023-12-06
Vera/ENG-463 Empty Data Bugfix

recursivepaws:
# Description
During the wasm mounting stage, banyan-cli was creating a new mount and pushing it if it failed to pull one. 
This was incorrect, and so was removed. 
That said, we still needed a way to perform `WasmMount::new` on drive creation. A wasm mount will fail to pull if it has not been initialized, but this is not definitive indication that it wasnt initialized. 

To accomplish this, `create_bucket` was refactored to `create_bucket_and_mount` (`createBucketAndMount`) which performs the necessary mounting initialization in tandem with the bucket creation, returning a simple `WasmBucketMount` type with getters for both internal fields.
Another minor fix was also implemented to fix an error in the `write_read` wasm test case, and wasm testing was modified to fit this new bucket creation paradigm.

## Link to issue
https://linear.app/banyan/issue/ENG-463/client-overwriting-buckets-w-empty-data

## Link to corresponding PR
https://github.com/banyancomputer/banyan-core/pull/252

## Type of change


src/wasm/compat/types/mount.rs
```diff
@@ -621,15 +625,19 @@ impl WasmMount {
...
+            api_blockstore
+                .find_cids(cids)
+                .await
+                .map_err(to_wasm_error_with_msg("find_cids"))?;
         }
 
         info!("read_bytes() running fs.read @ {:?}", path_segments);
 
+        // Attempt to fetch from local first, remote second
+        let split_store = DoubleSplitStore::new(&self.content_blockstore, &api_blockstore);
```
recursivepaws, line 638: this fixes the `write_read` test in web.rs and i think is a good idea to fallback on in our memory `BlockStore` first before seeking the network's help. that said the fact that the `write_read` test fails without this implementation makes me worry that something is still wrong with the api blockstore.

sstelfox: If things are intermittently failing for a test like this, I would guess its because its not waiting for the upload to get applied after the data upload finishes. There could definitely be more bugs in the implementation, but this kind of test that relies on remote services (even ours) aren't great tests to rely because we can't reason about the asynchronous operations behind the scenes or whether the remote service is behaving correctly.

In our case, after the data is uploaded to whatever service acts as a storage provider, there is an additional check-in that needs to happen between the storage provider and the core services that happens automatically. That check-in may not start until 500ms after the upload finishes based on a variety of factors in the storage provider service. The check-in both registers the locations of the blocks that were just uploaded to the core service, and brings the just uploaded metadata to what we consider "current".

Until that happens the core service won't know about the blocks and will return a 404, and this is a great example of why network requests should be generally avoided in unit tests. This PR doesn't need to correct the tests in any way, I just saw it as an opportune example for some exposition.

sstelfox (approved):

PR #311, opened 2023-12-11
Vera/ENG-470 Unblocking Alex

recursivepaws:
# Description
Alex has pointed out two glaring annoyances to me which have been preventing him from completing his work on the receive file ticket. He wants to write a native test for receiving a file, but:
1. There is no longer a functional integration for the CLI and frontend
2. Setting the api endpoint using CLI commands almost always fails

These changes exist to address these issues. 
- Refactored wasm function `complete_device_registration` -> `approve_device_api_key` (`approveDeviceApiKey`)
    - Since the loss of the device API key registration route in Next.js, connecting the CLI with the frontend has remained broken in the background while other tickets have taken priority. 
    - Internally, the new version of this command calls `CreateDeviceApiKey`, using the PEM of the PublicKey associated with the device, which is likewise passed in a URL-encoded base64 string to the new `/register-device/:spki` endpoint on core. After this call succeeds, it calls `EndRegwait` using the fingerprint that `CreateDeviceApiKey` returned, which should match that which is already on the event bus assuming everything has gone smoothly up until this point. 
    - This, along with [the corresponding change on core](https://github.com/banyancomputer/banyan-core/pull/270), updates makes this integration functional again. 
- Restructured `RunnableCommand`, `GlobalConfig`, and `OmniBucket` functions to simplify the lifetimes of data associated with an individual CLI run. 
    - The previous version of this trait implementation required `&mut` handles on the `GlobalConfig` and `Client`. 
        - Even though everything should have been working as expected, I believe this tripped up the borrow checker and lead to issues with the wrong versions of these structs being serialized when the `run_internal` call was completed. 
        - This resulted in the very frustrating outcome of having some CLI commands successfully update the global config.json state, while other failed to entirely (notably the API endpoint command). 
        - This change fixes that bug. 
    - This also has the added benefit of allowing for far more succinct code across the board in the CLI submodule. 
    - `GlobalConfig` functions which self-modify also save to disk now. The need to do so independently of updating this data structure was causing ambiguity in the source of truth. 

## Link to issue
https://linear.app/banyan/issue/ENG-470/unblocking-alex

## Type of change

### DIFF
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -1,3 +1,5 @@
+use crate::native::configuration::globalconfig::GlobalConfig;
+
 use super::{
     error::ApiError,
     requests::{ApiRequest, StreamableApiRequest},
@@ -139,6 +141,13 @@ impl Client {
                 match &self.signing_key {
                     Some(signing_key) => {
                         self.bearer_token = Some(claims.encode_to(signing_key).await?);
+
+                        #[cfg(not(target_arch = "wasm32"))]
+                        {
+                            let mut global = GlobalConfig::from_disk().await.unwrap();
+                            global.save_client(self.clone()).await.unwrap();
+                        }
+
                         Ok(self.bearer_token.clone().unwrap())
                     }
                     _ => Err(ApiError::auth_required()),
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -147,8 +147,13 @@ impl Client {
                         #[cfg(not(test))]
                         #[cfg(not(target_arch = "wasm32"))]
                         {
-                            let mut global = GlobalConfig::from_disk().await.unwrap();
-                            global.save_client(self.clone()).await.unwrap();
+                            let mut global = GlobalConfig::from_disk()
+                                .await
+                                .map_err(|_| ApiError::update_client())?;
+                            global
+                                .save_client(self.clone())
+                                .await
+                                .map_err(|_| ApiError::update_client())?
                         }
 
                         Ok(self.bearer_token.clone().unwrap())
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -1,3 +1,5 @@
+#[cfg(not(test))]
+#[cfg(not(target_arch = "wasm32"))]
 use crate::native::configuration::globalconfig::GlobalConfig;
 
 use super::{
@@ -142,6 +144,7 @@ impl Client {
                     Some(signing_key) => {
                         self.bearer_token = Some(claims.encode_to(signing_key).await?);
 
+                        #[cfg(not(test))]
                         #[cfg(not(target_arch = "wasm32"))]
                         {
                             let mut global = GlobalConfig::from_disk().await.unwrap();
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -1,7 +1,3 @@
-#[cfg(not(test))]
-#[cfg(not(target_arch = "wasm32"))]
-use crate::native::configuration::globalconfig::GlobalConfig;
-
 use super::{
     error::ApiError,
     requests::{ApiRequest, StreamableApiRequest},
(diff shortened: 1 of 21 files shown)


src/api/client.rs
```diff
@@ -139,6 +143,19 @@ impl Client {
                 match &self.signing_key {
                     Some(signing_key) => {
                         self.bearer_token = Some(claims.encode_to(signing_key).await?);
+
+                        #[cfg(not(test))]
```
amiller68: Why are you saving a reqwest client to disk

src/cli/commands/account.rs
```diff
@@ -74,13 +74,24 @@ impl RunnableCommand<NativeError> for AccountCommand {
...
-                .expect("failed to open browser");
+                // URL encoded DER bytes
+                let spki_b64 =
+                    general_purpose::STANDARD.encode(public_device_key.export_bytes().await?);
+                let spki_b64_url_safe =
+                    url::form_urlencoded::byte_serialize(spki_b64.as_bytes()).collect::<String>();
+                // Construct the proper URL to open
+                let url = global
+                    .get_endpoint()
+                    .join(&format!("register-device/{}", spki_b64_url_safe))
```
amiller68, line 85: this is not an endpoint we support

amiller68: we do have this but its authenticated route.
https://github.com/banyancomputer/banyan-core/blob/main/crates/banyan-core-service/src/api/auth/create_device_api_key.rs

src/cli/commands/drives.rs
```diff
@@ -72,15 +70,11 @@ pub enum DrivesCommand {
...
-    async fn run_internal(
-        self,
-        global: &mut GlobalConfig,
-        client: &mut Client,
-    ) -> Result<String, NativeError> {
+    async fn run_internal(self) -> Result<String, NativeError> {
         match self {
             // List all Buckets tracked remotely and locally
             DrivesCommand::Ls => {
-                let omnis = OmniBucket::ls(global, client).await;
```
amiller68, line 83: whats up with this interface change

src/cli/commands/account.rs
amiller68: oh i see yes theres soemthing in the FE for this

src/api/client.rs
amiller68: Yeah you should not be saving this information everytime u generate a bearer token, would appreciate context on the motivation behind this if this is necessary

recursivepaws: its not actually saving the whole client object to disk, internally it's saving the api key as a pem and the user-id to the global config. 
this only ever runs when the signing key needs to be updated, which is also the context in which we would want to re-serialize it anyway. 
the benefit of embedding it in this function is not needing to serialize the client to global config state *every time* we do an API call "just in case" this key got updated. instead, this serialization only occurs when a change actually happens during a `Client` call.

src/cli/commands/account.rs
recursivepaws: https://github.com/banyancomputer/banyan-core/pull/270

src/cli/commands/drives.rs
recursivepaws: passing these `&mut` pointers around is why the wrong version of a GlobalConfig ended up being serialized in `run()`. This is far more atomic and predictable- read the PR description for further details

src/api/client.rs
recursivepaws: as a motivation reminder- not serializing this means we'd be calling the `refresh()` function presumably every time start the cli even if we still had a valid token. useless api calls bad. we could reduce the frequency here even further but it would have its own drawback

amiller68: all the information needed to generate this bearer token should be on disk already.

this does not run only when the signing key is updated -- it runs on every request

we should never have to sererialize or deserialize this client, the parameters to build it should be configurable

everything you need to generate this token (uuid, ecdsa key, remote config) should be things you can configure at runtime / pull from disk

i dont think i can put a check on this, all this function should do is return a string, and not generate side effects

src/cli/commands/drives.rs
amiller68: what im getting at is i asked to be unblocked on login, this change couldve been encapulsated in another ticket . change. its fine but just unexepcted

src/native/configuration/globalconfig.rs
```diff
@@ -113,6 +113,7 @@ impl GlobalConfig {
     }
 
     /// Save the Client data to the config
+    #[allow(unused)]
     pub async fn save_client(&mut self, client: Client) -> Result<(), NativeError> {
```
amiller68: the client should not populate the config

the config should populate the client

amiller68 (changes requested):
Overall fine, but bits about client serializtion need to be taken out

amiller68 (approved):
👍

PR #314, opened 2023-12-12
Alex/sharing link api

amiller68:
# Description
Fixes sharing by:

In Shared Code
1. ensuring that file references get stored in content on write
2. ensuring that file references get stored in content on mv
3. ensuring that snapshot share pointer get stored in content on share_file

In Wasm Code
4. making share an `append` operation
5. making mv an `append` operation

Why:
- There's no way to get around item 3
- There should be a way to get around items 1 & 2 & 5 by writing said blocks explicitly to the content store. But for some reason the PrivateFile refs don't persist in wasm, even if using the same content blockstore. This needs more investigation but I have found no pattern including PrivateFile::store other than the one presented here that works correctly.

Tl;Dr

Sharing works now, but this makes share_file and mv append operations in the wasm client

### DIFF
--- a/tests/web.rs
+++ b/tests/web.rs
@@ -187,7 +187,9 @@ mod test {
             )
             .await?;
 
-        let shared_file = fs_metadata.share_file(&cat_path, &content_store).await?;
+        let shared_file = fs_metadata
+            .share_file(&cat_path, &metadata_store, &content_store)
+            .await?;
 
         let share_string = shared_file.export_b64_url()?;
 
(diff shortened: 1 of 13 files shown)


tests/web.rs
```diff
@@ -164,6 +166,118 @@ mod test {
...
+    //         .read_shared_file(shared_string)
+    //         .await
+    //         .expect("read_shared_file_from_bs failed")
+    //         .to_vec();
+
+    //     // Assert successful reconstruction
+    //     assert_eq!(new_bytes, zero_content_buffer.to_vec());
+
+    //     Ok(())
+    // }
```
sstelfox, line 279: Talked to Alex about these commented out tests, it seems they're likely failing due to the outstanding metadata bug and should be re-enabled when that has been resolved.

sstelfox (approved):

PR #319, opened 2023-12-21
Vera/ENG-474 Block Writing API

recursivepaws:
# Description
Added new Api methods and test case for writing individual Blocks to the staging service, as well as retrieving those blocks written. 

## Links
https://github.com/banyancomputer/banyan-core/pull/293
https://linear.app/banyan/issue/ENG-474/block-writing-api


sstelfox:
I mentioned this in a prior meeting, but I do not want to merge this. The ticket was specifically only for the server side. I don't want to change anything about the client behavior around storage until we remove WNFS which will invalidate a lot of this PR.

PR #323, opened 2024-02-01
Expose Remount / Bucket Refresh to WASM Clients

sstelfox:
Exposes as `remount()` function on WasmMount objects that will pull the most recent metadata from the server and unlocking it in place to act as a bandaid for the 409 Conflict problem.

I'm not sure why the compiler started detecting a couple of the response types as dead code, I did confirm they were still present in code paths (it might be from different architectures).

I also tweaked the logging to give us a bit more detail right now of whats going on. It might give us better insights into the problem.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1400,9 +1400,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.3.22"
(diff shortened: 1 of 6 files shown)


recursivepaws (approved):
lgtm

PR #325, opened 2024-02-23
ENG-561 Issue Renaming Buckets

recursivepaws:
Tiny modification to the api submodule which alters the api request sent on bucket updating to match that of the core service as per [this ticket](https://linear.app/banyan/issue/ENG-561/issue-renaming-buckets).

### DIFF
--- a/src/api/requests/core/buckets/update.rs
+++ b/src/api/requests/core/buckets/update.rs
@@ -2,11 +2,13 @@ use std::error::Error;
 use std::fmt::{self, Display, Formatter};
 
 use reqwest::{Client, RequestBuilder, Url};
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 
 use crate::api::requests::ApiRequest;
 
+#[derive(Serialize)]
 pub struct UpdateBucket {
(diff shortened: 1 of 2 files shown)


src/api/requests/core/buckets/update.rs
```diff
@@ -17,9 +20,10 @@ impl ApiRequest for UpdateBucket {
 
     fn build_request(self, base_url: &Url, client: &Client) -> RequestBuilder {
         let url = base_url
-            .join(&format!("/api/v1/buckets/{}", self.0.id))
+            .join(&format!("/api/v1/buckets/{}", self.bucket_id))
             .unwrap();
-        client.put(url).json(&self.0)
+
+        client.put(url).json(&self.name)
```
sstelfox: Per the other PR, this does need to be a structure JSON request and while a string is technically a valid JSON construction it's really bad practice to JSON encoded a lone string

sstelfox (changes requested):
This seems ok... But I'm not sure how the associated

sstelfox (approved):

PR #327, opened 2024-06-03
Vera/ENG-617 banyanfs rewrite

recursivepaws:
# Description
Rewrote the entire repo, yay!
- Pretty table output for CLI
- banyanfs integration
- removed WNFS, CAR, etc. 
- pruned 454 dependencies down to 303
  - no longer exposed to any security vulnerabilities, audits pass w flying colors
- changed to a binary from a workspace
- implemented `OnDisk` type for neatly managing user resources in an XDG compliant way while minimizing programmer overhead on individual resource types
- created implementations for various banyanfs types which enable us to store data on disk across user sessions rather than solely ephemerally / in memory
- created resilient system for managing drive and key states even when a user signs out or loses network connectivity
- rewrote continuous integration testing to be congruent with standards set in our core repo
- created a UX of minimal overhead. no syncing, uploading, downloading, etc. that all gets handled for you. you either prepare your local drive, or restore it from the network.
  - this will become even more seamless with the integration of fuse
  - updated README to make it not wrong even if it isn't perfect
  
known TODOs:
- progress bars. i really like the way [kdam](https://crates.io/crates/kdam) can be barebones or fancy, and is actively maintained. right now, i cant peer into banyanfs operations. implementing feels like a waste when we only have file level progress increments.
- files have to be loaded entirely into memory before being written, when they are downloaded / restored. more efficient way to do this?


src/cli/mod.rs
```diff
@@ -1,157 +1,39 @@
...
+#[derive(clap::Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+pub struct Args {
+    /// Command passed
+    #[command(subcommand)]
+    pub command: commands::BanyanCommand,
+    /// Verbosity level.
+    #[arg(short, long, help = "verbosity level", default_value = "normal")]
+    pub verbose: verbosity::MyVerbosity,
+}
```
jscatena88: I may be just be missing it it but it looks like the verbosity switch isn't wired up to anything

recursivepaws: nice catch yeah, more vestigial code

recursivepaws:
some of these CI should be `required`, the `required` ones dont exist anymore, i dont have settings access on this repo i assume that's where they are

